PR #11, opened 2015-02-18
Use urllib for better OAuth2 URL construction

jmoldow:
Use ``six.moves.urllib.parse.urlencode()`` to correctly encode the query string parameters that make up the OAuth2 authorize URL.

Previously, the URL was constructed via direct string formatting. This meant that if reserved or non-ASCII characters were present in the data, they would be used unencoded in the URL. This was particularly
problematic with the ``redirect_uri`` query string parameter:

- It is itself a URI and therefore likely to contain reserved characters.

- It is user-supplied and therefore has the possibility of containing non-ASCII unicode characters.

The correct encoding of characters in URLs is specified in various Internet RFCs, which the ``urllib`` package has been designed to match. So using ``urllib`` fixes these problems.

This commit also corrects some related issues:

- When the ``redirect_url`` parameter is not supplied to ``OAuth2.get_authorization_url()``, the ``redirect_uri`` query string parameter is omitted from the constructed URL.

- Text is utf-8 encoded to byte strings before being URL encoded.

- Fixes a misuse of the output of ``six.moves.urllib.parse.parse_qs()``. The values of the dictionary were being assumed to be strings, when they are actually lists of strings. And in Python 2, these strings need to be manually decoded from ``bytes`` to ``unicode``.

- Adds unit test coverage for non-ASCII and ``None`` values of the ``redirect_url`` parameter to ``OAuth2.get_authorization_url()``.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -6,6 +6,8 @@ from threading import Lock
 import random
 import string  # pylint:disable=deprecated-module
 
+from six.moves.urllib.parse import urlencode, urlunsplit  # pylint:disable=import-error,no-name-in-module
+
 from boxsdk.network.default_network import DefaultNetwork
 from boxsdk.config import API
 from boxsdk.exception import BoxOAuthException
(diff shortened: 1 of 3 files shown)


boxcla:

Jeff-Meadows:
:+1:

PR #17, opened 2015-02-25
Use the built-in enumerate function instead of maintaining a separate counter

aptxkid:


boxcla:

Jeff-Meadows:
Can you amend the commit message to be more specific about where you're doing this?

Otherwise, LGTM :+1:

PR #20, opened 2015-03-01
Add support for metadata.

### LINKED ISSUE #16: Add support for Box metadata

Jeff-Meadows:
Fixes #16.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -181,10 +181,10 @@ Metadata
     client.file(file_id='SOME_FILE_ID').metadata().create({'key', 'value')
 
     # Update metadata
-    from boxsdk.object.metadata import MetadataUpdate
-    update = MetadataUpdate()
+    metadata = client.file(file_id='SOME_FILE_ID').metadata()
+    update = metadata.start_update()
     update.add('/key', 'new_value')
-    client.file(file_id='SOME_FILE_ID').metadata().update(update)
+    metadata.update(update)
 
 
 Contributing
--- a/README.rst
+++ b/README.rst
@@ -169,6 +169,23 @@ Events
     # Get latest stream position
     client.events().get_latest_stream_position()
 
+Metadata
+~~~~~~~~
+
+.. code-block:: python
+
+    # Get metadata
+    client.file(file_id='SOME_FILE_ID').metadata().get()
+
+    # Create metadata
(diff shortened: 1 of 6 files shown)


boxcla:

README.rst
```diff
@@ -169,6 +169,23 @@ Events
...
+.. code-block:: python
+
+    # Get metadata
+    client.file(file_id='SOME_FILE_ID').metadata().get()
+
+    # Create metadata
+    client.file(file_id='SOME_FILE_ID').metadata().create({'key', 'value')
+
+    # Update metadata
+    from boxsdk.object.metadata import MetadataUpdate
```
aptxkid: Should we start putting classes for external access to the __init__ .py of the boxsdk package? Then users can do `from boxsdk import MetadataUpdate`.

boxsdk/object/file.py
```diff
@@ -1,10 +1,9 @@
 # coding: utf-8
 
 from __future__ import unicode_literals
-
 from boxsdk.config import API
-
-from .item import Item
+from boxsdk.object.item import Item
```
aptxkid: I think we should at least be consistent about if we are going to import with full path or relative path.

boxsdk/object/metadata.py
```diff
@@ -0,0 +1,187 @@
...
+        :type value:
+            `unicode`
+        :param old_value:
+            If specified, only update the key/value pair if the current value is equal to oldValue.
+        :type old_value:
+            `unicode`
+        """
+        if old_value is not None:
+            self._ops.append({'op': 'test', 'path': path, 'value': old_value})
+        self._ops.append({'op': 'replace', 'path': path, 'value': value})
```
aptxkid: I guess you can merge all these operation methods into one.

aptxkid:
Added some minor comments. Otherwise, LGTM :+1:

boxsdk/object/file.py
Jeff-Meadows: Yep. I think we should use absolute, but we can update that in a different PR.

README.rst
Jeff-Meadows: Yeah, it shouldn't be so hard to do an update. Either from boxsdk import MetadataUpdate, or, perhaps better, add a method to Metadata that will give you a MetadataUpdate object.

PR #21, opened 2015-03-01
Improve demo code by making it more robust.

### LINKED ISSUE #7: Make demo program more robust
The demo program ought to adhere t0
1) It leaves the target acct untouched (aka demo cleans up after itself)
2) If a target acct doesn't support some functionality, then the demo program gracefully avoids that functionality.
3) If demo program is interrupted, and thus leaves "turds" in the target acct... a subsequent run of the demo program should not barf. It should gracefully clean up that earlier stuff.

### LINKED ISSUE #10: Demo program incompatible with Python 2.6, Python 3

Jeff-Meadows:
Also support py2.6, py3, and pypy with the demo.

Fixes #7 Fixes #10

### DIFF
--- a/demo/auth.py
+++ b/demo/auth.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-from __future__ import unicode_literals
+from __future__ import print_function, unicode_literals
 
 import bottle
 import os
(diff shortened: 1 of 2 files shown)


boxcla:

aptxkid:
LGTM :+1:

PR #25, opened 2015-03-05
Add preflight

aptxkid:
Add preflight interface to `File` class.

I will add preflight to file update and file upload in later PRs.

### DIFF
--- a/boxsdk/object/base_endpoint.py
+++ b/boxsdk/object/base_endpoint.py
@@ -17,8 +17,7 @@ class BaseEndpoint(object):
         """
         self._session = session
 
-    @classmethod
-    def get_url(cls, endpoint, *args):
+    def get_url(self, endpoint, *args):
         """
         Return the URL used to access the endpoint.
 
--- a/boxsdk/object/base_endpoint.py
+++ b/boxsdk/object/base_endpoint.py
@@ -17,7 +17,8 @@ class BaseEndpoint(object):
         """
         self._session = session
 
-    def get_url(self, endpoint, *args):
+    @classmethod
+    def get_url(cls, endpoint, *args):
         """
         Return the URL used to access the endpoint.
 
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -65,14 +65,14 @@ class BaseObject(BaseEndpoint):
         """
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
-    def get_type_url(self, *args):
+    def get_type_url(self):
         """
-        Return the URL for type of the given resource, appending any optional parts as specified by args.
+        Return the URL for type of the given resource.
 
         :rtype:
             `unicode`
         """
-        return super(BaseObject, self).get_url('{0}s'.format(self._item_type), *args)
+        return super(BaseObject, self).get_url('{0}s'.format(self._item_type))
 
     @property
     def object_id(self):
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -65,14 +65,15 @@ class BaseObject(BaseEndpoint):
         """
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
-    def get_type_url(self):
+    @classmethod
+    def get_type_url(cls, *args):
         """
-        Return the URL for type of the given resource.
+        Return the URL for type of the given resource, appending any optional parts as specified by args.
 
         :rtype:
             `unicode`
         """
-        return super(BaseObject, self).get_url('{0}s'.format(self._item_type))
+        return super(BaseObject, cls).get_url('{0}s'.format(cls._item_type), *args)
 
     @property
     def object_id(self):
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -65,15 +65,14 @@ class BaseObject(BaseEndpoint):
         """
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
-    @classmethod
-    def get_type_url(cls, *args):
+    def get_type_url(self, *args):
         """
         Return the URL for type of the given resource, appending any optional parts as specified by args.
 
         :rtype:
             `unicode`
         """
-        return super(BaseObject, cls).get_url('{0}s'.format(cls._item_type), *args)
+        return super(BaseObject, self).get_url('{0}s'.format(self._item_type), *args)
 
     @property
     def object_id(self):
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -86,7 +86,7 @@ class File(Item):
         for chunk in box_response.network_response.response_as_stream.stream(decode_content=True):
             writeable_stream.write(chunk)
 
-    def update_contents_with_stream(self, file_stream, preflight_check=False, etag=None):
+    def update_contents_with_stream(self, file_stream, etag=None):
         """
         Upload a new version of a file, taking the contents from the given file stream.
 
@@ -94,11 +94,6 @@ class File(Item):
             The file-like object containing the bytes
         :type file_stream:
             `file`
-        :param preflight_check:
-            Whether or not makes an extra API call before the update to check if the new file stream can be uploaded.
-            (https://developers.box.com/docs/#files-preflight-check)
-        :type preflight_check:
-            `bool`
         :param etag:
             If specified, instruct the Box API to update the item only if the current version's etag matches.
         :type etag:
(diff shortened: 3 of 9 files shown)


boxcla:

kapitoshka1812:
This looks good.

aptxkid:
I am working on unittest.

demo/example.py
```diff
@@ -4,6 +4,7 @@
 import os
 from boxsdk.client import Client
 from boxsdk.object.collaboration import CollaborationRole
+from boxsdk.object.file import File
```
Jeff-Meadows: Is this the only change to example.py?

boxsdk/object/base_endpoint.py
```diff
@@ -17,7 +17,8 @@ def __init__(self, session):
         """
         self._session = session
 
-    def get_url(self, endpoint, *args):
+    @classmethod
```
Jeff-Meadows: I don't think it's a good idea to make this a classmethod when subclasses will not make it a classmethod.

boxsdk/object/base_object.py
```diff
@@ -65,14 +65,15 @@ def get_url(self, *args):
         """
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
-    def get_type_url(self):
+    @classmethod
```
Jeff-Meadows: ditto

demo/example.py
aptxkid: @Jeff-Meadows I tested in example.py briefly. I will remove this line.

boxsdk/object/base_endpoint.py
aptxkid: hmm.. @Jeff-Meadows I agree that get_url() should not be a class method.

boxsdk/object/base_object.py
```diff
@@ -65,14 +65,14 @@ def get_url(self, *args):
         """
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
-    def get_type_url(self):
+    def get_type_url(self, *args):
```
aptxkid: I need to remove this change to get_type_url since it's no longer used.

boxsdk/object/file.py
```diff
@@ -11,6 +12,8 @@ class File(Item):
 
     _item_type = 'file'
 
+
```
aptxkid: Remove the empty lines.

boxsdk/object/item.py
```diff
@@ -5,11 +5,56 @@
 import json
 
 from .base_object import BaseObject
+from boxsdk.config import API
 
 
 class Item(BaseObject):
     """Box API endpoint for interacting with files and folders."""
 
+    @classmethod
```
kunalparmar: Why is this a class method?

aptxkid: Preflight is a little different than other API calls in a sense that you can do preflight when trying to update an existing file on Box or uploading a new file. So it doesn't make sense to make a normal method tie to an object. But it can be a staticmethod, if that's why you're asking.

kunalparmar: Is uploading a new file a class/static method?

boxsdk/object/item.py
```diff
@@ -5,11 +5,56 @@
 import json
 
 from .base_object import BaseObject
+from boxsdk.config import API
```
kunalparmar: Import style is inconsistent with the line above

boxsdk/object/item.py
```diff
@@ -5,11 +5,56 @@
...
+        url = '{0}/files/{1}'.format(API.BASE_API_URL, endpoint)
+        data = {'size': size}
+        if name:
+            data['name'] = name
+        if parent_id:
+            data['parent'] = {'id': parent_id}
+
+        session.options(
+            url=url,
+            expect_json_response=False,
```
kunalparmar: I don't see this called out in the developer docs. What does the options call return?

boxsdk/object/item.py
aptxkid: IMO, this is actually OK. Since 'item' module is in the 'object' package while apparently config module is not. Ideally, IMO, we would add 'from .config import API' to boxsdk/__init__.py and we can just do 'from boxsdk import API' here.

test/unit/object/test_item.py
```diff
@@ -105,3 +108,49 @@ def test_get(test_item_and_response, mock_box_session, fields, mock_object_id, e
...
     assert info.id == mock_object_id
+
+
+@pytest.mark.parametrize(
+    'size, name, file_id, parent_id, expected_url, expected_data',
+    [
+        (
+            # Testing preflight check of a new file to be uploaded to Box.
+            100,
+            'foo.txt',
```
kunalparmar: Rotunicode?

test/unit/object/test_item.py
```diff
@@ -105,3 +108,49 @@ def test_get(test_item_and_response, mock_box_session, fields, mock_object_id, e
...
+            100,
+            None,
+            '123456',
+            None,
+            '{0}/files/123456/content'.format(API.BASE_API_URL),
+            json.dumps({'size': 100}),
+        ),
+    ],
+)
+def test_preflight_check(
```
kunalparmar: You are missing tests for input combinations that may be None.

boxsdk/object/item.py
aptxkid: Uploading a file is a method of a File object, so it's tied to a file. One option would be making one preflight method to File for new file uploads, and adding one different preflight method to File for updating existing files on Box. In that way, nothing needs to be a class/static method. But we have two preflight methods. But that might be a good idea actually. Since code duplication can be addressed by putting code needs to be shared in Item. I think I like this idea more!

boxsdk/object/item.py
aptxkid: All we care about is if this call returns 200 or not. If it's not a 200 return, by default, the BoxSession will raise a BoxAPIException.

test/unit/object/test_item.py
aptxkid: We are simply checking if we are making the call with the same name. But sure! :)

test/unit/object/test_item.py
aptxkid: I believe these are all the combinations that Box supports. I can double check.

boxsdk/object/item.py
kunalparmar: You can do a relative import as from ..config import API

boxsdk/object/item.py
kunalparmar: Is this required then?

boxsdk/object/item.py
kunalparmar: That sounds like a more consistent design choice to me.

Jeff-Meadows: Yes, +1 to this.

test/unit/object/test_file.py
```diff
@@ -99,3 +99,43 @@ def test_unlock(test_file, mock_box_session, mock_file_response):
...
+        kwargs['name'] = name
+    test_file.preflight_check(**kwargs)
+    mock_box_session.options.assert_called_once_with(
+        url='{0}/files/{1}/content'.format(
+            API.BASE_API_URL,
+            mock_object_id,
+        ),
+        expect_json_response=False,
+        data=expected_data,
+    )
```
Jeff-Meadows: add newline

test/unit/object/test_folder.py
```diff
@@ -167,3 +167,19 @@ def test_update_sync_state(test_folder, mock_folder_response, mock_box_session,
     mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(data), params=None, headers=None)
     assert isinstance(update_response, Folder)
     assert update_response.object_id == test_folder.object_id
+
+
+def test_preflight(test_folder, mock_object_id, mock_box_session):
```
Jeff-Meadows: Parametrize file size/name.

boxsdk/object/item.py
```diff
@@ -5,11 +5,51 @@
...
+        """
+        Make an API call to check if certain file can be uploaded to Box or not.
+        (https://developers.box.com/docs/#files-preflight-check)
+
+        :param size:
+            The size of the file in bytes. Specify 0 for unknown file-sizes.
+        :type size:
+            `int`
+        :param name:
+            The name of the file to be uploaded. This is optional if `file_id` is specified.
```
Jeff-Meadows: end line with comma and start next line with lowercase but

boxsdk/object/item.py
```diff
@@ -5,11 +5,51 @@
...
 class Item(BaseObject):
     """Box API endpoint for interacting with files and folders."""
 
+    def _preflight_check(self, size, name=None, file_id=None, parent_id=None):
+        """
+        Make an API call to check if certain file can be uploaded to Box or not.
+        (https://developers.box.com/docs/#files-preflight-check)
+
+        :param size:
+            The size of the file in bytes. Specify 0 for unknown file-sizes.
```
Jeff-Meadows: The file to be uploaded.

Change to "unknown file size"

test/unit/object/test_folder.py
aptxkid: Any suggestions about exactly how to parametrize size/name? I did it for preflight in File because 'name' can be omitted.

test/unit/object/test_file.py
aptxkid: Done

Jeff-Meadows:
LGTM after travis is done.

PR #33, opened 2015-03-21
Add preflight option for updating files

aptxkid:
Add preflight option to update_contents_with_stream and update_contents,
so files can be updated with optional preflight check before actually
uploading all the bytes.

### DIFF
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -60,7 +60,7 @@ class File(Item):
         for chunk in box_response.network_response.response_as_stream.stream(decode_content=True):
             writeable_stream.write(chunk)
 
-    def update_contents_with_stream(self, file_stream, etag=None, preflight_check=False, size=0):
+    def update_contents_with_stream(self, file_stream, etag=None, preflight_check=False, preflight_expected_size=0):
         """
         Upload a new version of a file, taking the contents from the given file stream.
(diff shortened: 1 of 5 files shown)


boxcla:

boxsdk/object/file.py
```diff
@@ -72,12 +72,26 @@ def update_contents_with_stream(self, file_stream, etag=None):
             If specified, instruct the Box API to update the item only if the current version's etag matches.
         :type etag:
             `unicode` or None
+        :param preflight_check:
+            If specified, preflight check will be performed before actually uploading the file.
+        :type preflight_check:
+            `bool`
+        :param size:
```
Jeff-Meadows: I think it's unintuitive for there to be a separate parameter called size which is only relevant for preflight, yet the parameter name doesn't mention it. I would suggest either renaming size to expected_size_for_preflight or, alternatively, changing preflight_check's type to a dictionary that can optionally contain a size key. Then if preflight_check is None, skip the check. If it's not None, do the check, and see if there's a size key.

Jeff-Meadows:
This commit is missing tests for what happens if preflight fails:
* The file shouldn't be uploaded
* An exception should be raised

aptxkid:
LGTM :+1:

PR #37, opened 2015-03-25
Add support for searching by metadata value (Fixes #36)

### LINKED ISSUE #36: Add support for searching by metadata value
The Search API now has the ability to search by metadata value using the "mdfilters" param.  The SDK should have support for this new functionality.

hnguyen08:
Added mdfilters param to search object

### DIFF
--- a/boxsdk/client.py
+++ b/boxsdk/client.py
@@ -105,7 +105,7 @@ class Client(object):
         response = box_response.json()
         return [User(self._session, item['id'], item) for item in response['entries']]
 
-    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None, mdfilters=None):
+    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None, metadata_filters=None):
         """
         Search Box for items matching the given query.
 
@@ -124,21 +124,26 @@ class Client(object):
         :param ancestor_folders:
             Folder ids to limit the search to.
         :type ancestor_folders:
-            `list` of :class:`Folder`
+            `iterable` of :class:`Folder`
         :param file_extensions:
             File extensions to limit the search to.
         :type file_extensions:
-            `list` of `unicode`
-        :param mdfilters:
+            `iterable` of `unicode`
+        :param metadata_filters:
             Filters used for metadata search
-        :type mdfilters:
+        :type metadata_filters:
             :class:`MetadataSearchFilters`
         :return:
             A list of items that match the search query.
         :rtype:
             `list` of :class:`Item`
         """
-        return Search(self._session).search(query, limit, offset, ancestor_folders, file_extensions, mdfilters)
+        return Search(self._session).search(query=query,
+                                            limit=limit,
+                                            offset=offset,
+                                            ancestor_folders=ancestor_folders,
+                                            file_extensions=file_extensions,
+                                            metadata_filters=metadata_filters)
 
     def events(self):
         """
--- a/boxsdk/client.py
+++ b/boxsdk/client.py
@@ -105,7 +105,7 @@ class Client(object):
         response = box_response.json()
         return [User(self._session, item['id'], item) for item in response['entries']]
 
-    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None):
+    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None, mdfilters=None):
         """
         Search Box for items matching the given query.
 
@@ -124,17 +124,21 @@ class Client(object):
         :param ancestor_folders:
             Folder ids to limit the search to.
         :type ancestor_folders:
-            `list` of `unicode`
+            `list` of :class:`Folder`
         :param file_extensions:
             File extensions to limit the search to.
         :type file_extensions:
             `list` of `unicode`
+        :param mdfilters:
+            Filters used for metadata search
+        :type mdfilters:
+            :class:`MetadataSearchFilters`
         :return:
             A list of items that match the search query.
         :rtype:
             `list` of :class:`Item`
         """
-        return Search(self._session).search(query, limit, offset, ancestor_folders, file_extensions)
+        return Search(self._session).search(query, limit, offset, ancestor_folders, file_extensions, mdfilters)
 
     def events(self):
         """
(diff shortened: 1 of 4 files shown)


boxcla:
Verified that @hnguyen08 has signed the CLA. Thanks for the pull request!

boxsdk/client.py
```diff
@@ -105,7 +105,7 @@ def users(self):
         response = box_response.json()
         return [User(self._session, item['id'], item) for item in response['entries']]
 
-    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None):
+    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None, mdfilters=None):
```
aptxkid: The argument list is getting long. I would suggest break it into multiple lines. e.g.:
def search(
  self,
  query,
  ...
)

boxsdk/object/search.py
```diff
@@ -5,12 +5,147 @@
...
+        """
+        range_part = {}
+        if gt_value:
+            range_part['gt'] = gt_value
+        if lt_value:
+            range_part['lt'] = lt_value
+        self._field_filters.update({field_key: range_part})
+
+
+class MetadataSearchFilters(object):
```
aptxkid: I would suggest import MetadataSearchFilters in boxsdk/_ _ init _ _.py, since it would be used in the search API in client.

boxsdk/object/search.py
```diff
@@ -5,12 +5,147 @@
...
+            :class:`MetadataSearchFilter`
+        """
+        self._filters.append(filter)
 
 
 class Search(BaseEndpoint):
     """Search Box for files and folders."""
 
-    def search(self, query, limit, offset, ancestor_folders=None, file_extensions=None):
+    def get_url(self, *args):
```
Jeff-Meadows: Do you use this in the class itself?

boxsdk/client.py
Jeff-Meadows: I know it's called mdfilters in the API, but can you call the parameter metadata_filters?

boxsdk/client.py
```diff
@@ -124,17 +124,21 @@ def search(self, query, limit, offset, ancestor_folders=None, file_extensions=No
         :param ancestor_folders:
             Folder ids to limit the search to.
         :type ancestor_folders:
-            `list` of `unicode`
+            `list` of :class:`Folder`
```
Jeff-Meadows: Good catch!

And actually we could be more flexible and call it `Iterable` of folder.

boxsdk/object/search.py
```diff
@@ -5,12 +5,147 @@
...
 from boxsdk.util.translator import Translator
+import json
+
+class MetadataSearchFilter(object):
+    """
+    Helper class to encapsulate a single search filter. A search filter can only search against one template,
+    but can filter on many fields.
+    See :class:`MetadataSearchFilters`.
+    """
+    def __init__(self, template_key, scope):
```
Jeff-Meadows: In the rest of the SDK, we default these to 'properties' and 'global'. I think we should do that here, as well.

boxsdk/client.py
```diff
@@ -124,17 +124,21 @@ def search(self, query, limit, offset, ancestor_folders=None, file_extensions=No
...
+            Filters used for metadata search
+        :type mdfilters:
+            :class:`MetadataSearchFilters`
         :return:
             A list of items that match the search query.
         :rtype:
             `list` of :class:`Item`
         """
-        return Search(self._session).search(query, limit, offset, ancestor_folders, file_extensions)
+        return Search(self._session).search(query, limit, offset, ancestor_folders, file_extensions, mdfilters)
```
aptxkid: Break it into multiple lines and use kwargs here as well.

boxsdk/object/search.py
Jeff-Meadows: You just use search.start_mdfilters(). No need to import it into dunder init.

boxsdk/object/search.py
```diff
@@ -5,12 +5,147 @@
...
+        :type field_filter:
+            `unicode`
+        :param value:
+            The value to use to filter
+        :type value:
+            `unicode`
+        """
+        self._field_filters.update({field_key: value})
+
+    def add_range_filter(self, field_key, gt_value = None, lt_value = None):
```
Jeff-Meadows: What's the behavior if the user passes gt_value=None, lt_value=None?

Should we assert that at least one of the two is not None?

boxsdk/object/search.py
aptxkid: You are absolutely right.

boxsdk/object/search.py
```diff
@@ -5,12 +5,147 @@
...
+class MetadataSearchFilters(object):
+    """
+    Helper class to encapsulate a list of search filter params (mdfilters)
+    See https://developers.box.com/metadata-api/#search for more details
+    """
+    def __init__(self):
+        self._filters = []
+
+    @staticmethod
+    def make_single_mdfilter(template_key, scope):
```
aptxkid: To me, having a factory here makes the responsibility of this class unclear. How about getting rid of this factory method, and change the `add_filter` interface to accept `template_key` and `scope` so inside `add_filter` we can instantiate a MetadataSearchFilter instance and add that to the list.

aptxkid: To me, having this factory method in this MetadataSearchFilter collection class is a little weird.

aptxkid:
Looks very cool! Thanks, @hnguyen08

boxsdk/object/search.py
hnguyen08: Oh good point, I should just use it below where we construct the URL

boxsdk/object/search.py
hnguyen08: Unfortunately we cannot do filtered search on 'properties' metadata :(

boxsdk/object/search.py
hnguyen08: Yeah, I moved it to the Search class so all the factory methods are in there.

boxsdk/client.py
hnguyen08: When I break into multiple lines, pylint complains of duplicate-code (this file with object/search.py). It looks like I can't disable this rule locally.  Would you recommend disabling this rule globally or something else?

hnguyen08: Actually, the above comment was in reference to the method definition (search.py has the same argument list).  For now I will leave as single line.

PR #41, opened 2015-04-07
Add Box Accelerator Support

aptxkid:
Supports both new uploads and updating existing files on Box. I also
added a new example in demo/example.py.


boxcla:

kunalparmar:
:+1:

Jeff-Meadows:
I know you merged this already, but I think it could use some more work.

Correct me if I'm wrong, but I believe preflight + accelerator can be done using a single API call, but using the SDK we will always make separate API calls for each.

If so, let's open an issue to track combining the calls.

PR #46, opened 2015-05-05
Support unshared_at for getting shared links

### LINKED ISSUE #45: Add ability to set expiration date on shared links
`get_shared_link()` only takes access and etag as parameters. The create shared link API supports many other parameters, including unshared_at, permissions, can_download, can_preview.
https://box-content.readme.io/#page-create-a-shared-link-for-a-file

aptxkid:
This fixes #45

### DIFF
--- a/boxsdk/object/item.py
+++ b/boxsdk/object/item.py
@@ -159,7 +159,7 @@ class Item(BaseObject):
         }
         return self.update_info(data)
 
-    def get_shared_link(self, access=None, etag=None):
+    def get_shared_link(self, access=None, etag=None, unshared_at=None):
         """Get a shared link for the item with the given access permissions.
 
         :param access:
@@ -170,6 +170,11 @@ class Item(BaseObject):
         :param etag:
             If specified, instruct the Box API to create the link only if the current version's etag matches.
         :type etag:
+            `unicode` or None
+        :param unshared_at:
+            The day that this link should be disabled at. Timestamps are rounded off to the given day.
+            This field can only be set if the user is not a free user.
+        :type unshared_at:
             `unicode` or None
         :returns:
             The URL of the shared link.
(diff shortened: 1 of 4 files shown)


boxcla:

boxsdk/object/item.py
```diff
@@ -171,6 +171,11 @@ def get_shared_link(self, access=None, etag=None):
             If specified, instruct the Box API to create the link only if the current version's etag matches.
         :type etag:
             `unicode` or None
+        :param unshared_at:
+            The day that this link should be disabled at. Timestamps are rounded off to the given day.
+            This field can only be set if the user is not a free user.
+        :type unshared_at:
+            `unicode` or None
```
Jeff-Meadows: This is not Pythonic to take a unicode string. This should take a date or datetime.

Jeff-Meadows:
The issue specifically mentions unshared_at, but since the sharing API also allows you to set permissions, I think we should do that as well.

boxsdk/object/item.py
aptxkid: Good point. I was too lazy and trying to avoid the timezone. :p

test/unit/object/test_item.py
```diff
@@ -69,10 +91,27 @@ def test_get_shared_link(
     test_item, _ = test_item_and_response
     expected_url = test_item.get_url()
     mock_box_session.put.return_value.json.return_value = {'shared_link': {'url': test_url}}
-    url = test_item.get_shared_link(access, etag=etag)
+    expected_data = {'shared_link': {}}
+    if shared_link_access is not None:
+        expected_data['shared_link']['access'] = shared_link_access
```
aptxkid: How about using `@pytest.mark.parametrize` (in a `@genty`-ish way), so each expected_data can be paired with the input?

aptxkid:
:+1:  LGTM with one minor comment. :smiley:

PR #50, opened 2015-05-22
Create a convenience method that returns an iterator over the folder items

samkuehn:
This is equivalent to iterating over the folder item offsets manually.

I had trouble creating a test for this.  I understand if you don't want to merge this into the core SDK.  I find it useful so I thought I would share.

### DIFF
--- a/boxsdk/object/folder.py
+++ b/boxsdk/object/folder.py
@@ -154,6 +154,33 @@ class Folder(Item):
         response = box_response.json()
         return [Translator().translate(item['type'])(self._session, item['id'], item) for item in response['entries']]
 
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @samkuehn has signed the CLA. Thanks for the pull request!

HootyMcOwlface:
Consider using the _paging_wrapper in base_object.  Check out Group.membership().  Then you can build pagination into get_items.

potrebic:
yes, paging ought to be based on the pattern established in Group.membership (which leverages the reusable _paging_wrapper).

I suggest we closed the PR.

PR #51, opened 2015-05-28
Return bool from `Item.remove_shared_link()`.

jmoldow:
Make the return type match the type specified in the docstring.  The
method was previously attempting to return ``unicode or None``.

The method was also previously attempting to access the 'url' item on
the shared_link object. However, the Box API returns a null object when
the shared link is disabled [1]. Make the method check for success by
doing an ``is None`` check on the shared_link object.

[1] https://box-content.readme.io/#create-a-shared-link-for-a-folder


boxcla:

aptxkid:
LGTM :+1:

PR #52, opened 2015-06-17
Correct test invocations of Client.folder()

jmoldow:
Some tests were calling `Client.folder()` with a `folder_id` argument of 0. However, that parameter is supposed to be of type `unicode`, not `int`.

Correct all occurrences of `folder(0)` to `folder('0')`.


boxcla:

aptxkid:
:+1:  Good find!

Issue #47, opened 2015-05-11
Add seamless shared item headers


ianunruh:
To interact with shared items from the API, the `BoxApi: shared_link=SHARED_LINK` header must be included with requests.

My idea here would be to add a `_headers` dict to `BaseObject`. Then when `get_shared_item` is called with the share URL, it passes the header down to the `Folder` or `File` object. When requests are made on these items (`File#content`, `File#download_to`, `Folder#get_items`), they would use the `BoxApi` header and pass the header down recursively to subitems (when applicable).

Another, less generic solution would be to add `_shared_link` to `Item` and craft the headers from that.

Jeff-Meadows:
Hi Ian, I hope #54 provides a seamless experience with shared items. If not, feel free to reopen and explain why not!

PR #54, opened 2015-07-02
Add support for As-User and shared items headers.

Jeff-Meadows:
The Box API supports enterprise admins making requests as users via the As-User header.
It also supports gaining access to items via a shared link via the BoxApi: shared_link= header.

This commit adds an `as_user` and `with_shared_link` method to the SDK client, session, and endpoint objects
to support a fluent interface for adding the appropriate headers to requests.

For example, to get user with id 1234's root folder info:
    client.as_user(User('1234')).folder('0').get()

### DIFF
--- a/boxsdk/object/base_endpoint.py
+++ b/boxsdk/object/base_endpoint.py
@@ -50,7 +50,7 @@ class BaseEndpoint(object):
 
     def with_shared_link(self, shared_link, shared_link_password):
         """
-        Returns a new session object with default headers set up to make requests using the shared link for auth.
+        Returns a new endpoint object with default headers set up to make requests using the shared link for auth.
 
         :param shared_link:
             The shared link.
--- a/boxsdk/object/base_endpoint.py
+++ b/boxsdk/object/base_endpoint.py
@@ -36,3 +36,29 @@ class BaseEndpoint(object):
         url = ['{0}/{1}'.format(API.BASE_API_URL, endpoint)]
         url.extend(['/{0}'.format(x) for x in args])
         return ''.join(url)
+
+    def as_user(self, user):
+        """
+        Returns a new endpoint object with default headers set up to make requests as the specified user.
+
+        :param user:
+            The user to impersonate when making API requests.
+        :type user:
+            :class:`User`
+        """
+        return self.__class__(self._session.as_user(user))
+
+    def with_shared_link(self, shared_link, shared_link_password):
+        """
+        Returns a new session object with default headers set up to make requests using the shared link for auth.
+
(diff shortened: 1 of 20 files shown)


boxcla:

Jeff-Meadows:
Fixes #47

boxsdk/object/base_endpoint.py
```diff
@@ -36,3 +36,29 @@ def get_url(self, endpoint, *args):
...
+        :param user:
+            The user to impersonate when making API requests.
+        :type user:
+            :class:`User`
+        """
+        return self.__class__(self._session.as_user(user))
+
+    def with_shared_link(self, shared_link, shared_link_password):
+        """
+        Returns a new session object with default headers set up to make requests using the shared link for auth.
```
aptxkid: Typo new "endpoint" instead of "session"?

test/integration/test_as_user.py
```diff
@@ -0,0 +1,127 @@
...
+
+def test_client_as_user_causes_as_user_header_to_be_added(
+        box_client,
+        mock_box_network,
+        generic_successful_response,
+        mock_user_id,
+        box_api_headers,
+):
+    # pylint:disable=redefined-outer-name
+    mock_box_network.session.request.side_effect = [
```
aptxkid: Do you need the []? I remember for a single response you can just do `.side_effect  = return_value`

test/integration/test_as_user.py
```diff
@@ -0,0 +1,127 @@
...
+from __future__ import unicode_literals
+from mock import call, patch
+import pytest
+from boxsdk.config import API
+from boxsdk.object.group_membership import GroupMembership
+from boxsdk.object.user import User
+
+
+@pytest.fixture
+def box_api_headers(mock_user_id, access_token):
```
aptxkid: How about renaming this fixture to indicate that it has the as-user header?

aptxkid:
:+1:  LGTM with minor comments.

test/integration/test_as_user.py
Jeff-Meadows: You can do `.return_value = return_value`, but that's actually different than `.side_effect = [return_value]` because the return_value attribute does additional call tracking that I don't want in this test.

test/integration/test_as_user.py
Jeff-Meadows: Done.

boxsdk/object/base_endpoint.py
Jeff-Meadows: Done.

PR #63, opened 2015-08-04
Include count of items from Folder.get_items().

kunalparmar:
When paginating a folder's items, the number of items in the folder
could have changed between successive calls. Including the count of
items helps the caller adjust the pagination as required.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,7 +113,7 @@ Get items in a folder
 
 .. code-block:: python
 
-    items = client.folder(folder_id='0').get_items(limit=100, offset=0)
+    total_count, items = client.folder(folder_id='0').get_items(limit=100, offset=0)
 
 Create subfolder
 ~~~~~~~~~~~~~~~~
--- a/boxsdk/object/folder.py
+++ b/boxsdk/object/folder.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 import json
+from collections import namedtuple
 import os
 from six import text_type
 from boxsdk.config import API
@@ -123,6 +124,16 @@ class Folder(Item):
         """
         return self._get_accelerator_upload_url()
 
+    _FolderGetItemsResult = namedtuple(
+        'FolderGetItemsResult',
+        [
+            # Total number of items in the folder.
+            'total_count',
+            # List of items in the folder.
+            'items',
+        ],
+    )
+
     def get_items(self, limit, offset=0, fields=None):
         """Get the items in a folder.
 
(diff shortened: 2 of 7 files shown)


boxcla:
Verified that @kunalparmar has signed the CLA. Thanks for the pull request!

jmoldow:
This breaks backwards compatibility.

Can you explain the original motivation for making this change?

To me, based on the description given in the commit message, it doesn't seem like the use case is compelling enough to warrant a backwards incompatible change.

kunalparmar:
@jmoldow Currently, I would do something like this to list a folder -
```python
folder = client.folder(folder_id='0')
folder = folder.get()    # populate folder info
total_count = folder.item_collection['total_count']
listing = []
offset = 0
while offset < total_count:
    items = folder.get_items(limit=20, offset=offset)
    offset += len(items)
    listing += items
```

There were a few things I noticed in this -
1. I need an extra API call to get the count of items in the folder even though the API response from `get_items` has that data.
2. There is no way to short the while loop if the count of items in the folder has reduced as I'm iterating over it. This results in unnecessary API calls which can be avoided.

The proposed changes overcome these problems. With the changes I proposed, the code would be something like this -
```python
folder = client.folder(folder_id='0')
listing = []
offset = 0
total_count = 1
while offset < total_count:
    total_count, items = folder.get_items(limit=20, offset=offset)
    offset += len(items)
    listing += items
```

Jeff-Meadows:
I don't think we should merge this because of the fact that it breaks the existing API. @kunalparmar if you  feel strongly that the SDK should have this functionality, perhaps we could have a parallel ``get_items_with_count`` or similar.

potrebic:
@kunalparmar 
Putting aside the notion that the 'total_count' can change while paging... in the two code snippets I don't believe the logic for advancing **offset** was correct. 

```python
offset += len(items)
```
should be:
```python
offset += limit
```

The [V2 API for Get Items](https://box-content.readme.io/reference#get-a-folders-items) state:

> **offset**	StringURL Parameter (default=0)
> The offset at which to begin the response. An offset of value of 0 will start at the 
> beginning of the folder-listing. Note: If there are hidden items in your previous
> response, your next offset should be = offset + limit, not the # of records you
> received back.

potrebic:
@kunalparmar @Jeff-Meadows @jmoldow 

How about this idea... have folder.get_items() save the last value of 'total_count' in a public property on the folder object?

```python
...
while offset < total_count:
    items = folder.get_items(limit=20, offset=offset)
    total_count = folder.last_total_count
    ...
```

last_total_count could throw an exception if it is referenced before get_items is ever called.

Issue #65, opened 2015-08-21
Support setting passwords in get_shared_link()


Frencil:
The Box API [supports setting passwords on shared links](https://box-content.readme.io/reference#create-a-shared-link-for-a-file) but this feature is as yet missing from the shared link implementation in box-python-sdk ([here](https://github.com/box/box-python-sdk/blob/master/boxsdk/object/item.py#L162)).

I'm currently working up a pull request for this enhancement.

PR #66, opened 2015-08-21
Adding support for shared_link password parameter

Frencil:
Resolves Issue #65 by adding support for shared_link passwords.

### DIFF
--- a/boxsdk/object/item.py
+++ b/boxsdk/object/item.py
@@ -159,7 +159,7 @@ class Item(BaseObject):
         }
         return self.update_info(data)
 
-    def get_shared_link(self, access=None, etag=None, unshared_at=None, allow_download=None, allow_preview=None):
+    def get_shared_link(self, access=None, etag=None, unshared_at=None, password=None, allow_download=None, allow_preview=None):
         """Get a shared link for the item with the given access permissions.
 
         :param access:
@@ -176,6 +176,10 @@ class Item(BaseObject):
             and has permission to set expiration dates.
         :type unshared_at:
             :class:`datetime.date` or None
+        :param password:
+            The password required to view this link. If no password is specified, the default password will be used.
+        :type password:
+            `unicode` or None
         :param allow_download:
             Whether or not the item being shared can be downloaded when accessed via the shared link.
             If this parameter is None, the default setting will be used.
@@ -200,6 +204,9 @@ class Item(BaseObject):
 
         if unshared_at is not None:
             data['shared_link']['unshared_at'] = unshared_at.isoformat()
+
+        if password is not None:
+            data['shared_link']['password'] = password
 
         if allow_download is not None or allow_preview is not None:
             data['shared_link']['permissions'] = permissions = {}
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @Frencil, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Frencil:
CLA signed

boxcla:
Verified that @Frencil has just signed the CLA. Thanks, and we look forward to your contribution.

boxsdk/object/item.py
```diff
@@ -176,6 +176,10 @@ def get_shared_link(self, access=None, etag=None, unshared_at=None, allow_downlo
             and has permission to set expiration dates.
         :type unshared_at:
             :class:`datetime.date` or None
+        :param password:
+            The password required to view this link. If no password is specified, the default password will be used.
```
Jeff-Meadows: Is this correct? I thought that not specifying a password creates the shared link with no password.

Also, can you mention here that setting passwords is a premium feature, as is https://github.com/box/box-python-sdk/blob/master/boxsdk/object/folder.py#L192

boxsdk/object/item.py
```diff
@@ -159,7 +159,7 @@ def move(self, parent_folder):
         }
         return self.update_info(data)
 
-    def get_shared_link(self, access=None, etag=None, unshared_at=None, allow_download=None, allow_preview=None):
+    def get_shared_link(self, access=None, etag=None, unshared_at=None, password=None, allow_download=None, allow_preview=None):
```
Jeff-Meadows: I think that the password field should be added as the last parameter, so it doesn't break existing code that might pass all the parameters positionally (this would be a bad idea, but it would also be currently supported).

Jeff-Meadows:
LGTM with my 2 comments addressed.

Frencil:
Thanks for the feedback! FYI I see the Travis build failing just on the pypy environment because of this:

> ImportError: No module named cryptography.hazmat.bindings._constant_time

FWIW this was one error I also had locally and couldn't figure out a solution for. Since none of the changes in this branch affected that part of the test coverage I decided to proceed anyway, but if anyone knows how to fix that I'd love to get it fixed locally too.

Jeff-Meadows:
The PyPy failures look like a regression in the `cryptography` module. I've opened an issue with that project https://github.com/pyca/cryptography/issues/2290

Downgrading the package to version 0.9.3 with `.tox/pypy/bin/pip install cryptography==0.9.3 --upgrade` will allow the tests to pass locally.

boxsdk/object/item.py
```diff
@@ -176,6 +176,8 @@ def get_shared_link(self, access=None, etag=None, unshared_at=None, allow_downlo
             and has permission to set expiration dates.
         :type unshared_at:
             :class:`datetime.date` or None
+        :type password:
```
Jeff-Meadows: Last thing - these 2 lines need to move down in the docstring as well!

Frencil:
Not sure how I missed that last one, thanks. =)

PR #69, opened 2015-08-30
Identifier for the store tokens method

Tusky:
Adds ability to add an extra identifier, so it can be saved to the database with that. My change allows backwards compatibility, for those who doesn't need that feature.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -35,13 +35,13 @@ class OAuth2(object):
             Box API key used for identifying the application the user is authenticating with.
         :type client_id:
             `unicode`
-        :param record_identifier:
-            Local identifier for the store token method.
-        :type record_identifier:
-            `unicode`
         :param client_secret:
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @Tusky, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Tusky:
CLA signed

boxcla:
Verified that @Tusky has just signed the CLA. Thanks, and we look forward to your contribution.

jmoldow:
This change is unnecessary for the SDK (and also seems specific to whatever you're doing).

You can accomplish the same exact thing without an SDK change:

```python
from functools import partial

def store_tokens(access_token, refresh_token, record_identifier):
  ...

oauth = OAuth2(..., store_tokens=partial(store_tokens, record_identifier=u'identifier'))
```

Tusky:
Thanks, didn't know about partial and yeah in that case this is pointless :+1:

PR #70, opened 2015-09-11
Add 'kid' (jwt key id) to JWTAuth header

aptxkid:


boxcla:

jmoldow:
What is this? I don't see anything about this at https://developers.box.com/developer-edition/.

It sucks to not be able to test with the latest and greatest cryptography. It looks like Travis CI has an old version of PyPy. They have 2.5, but PyPy is up to 2.6. See https://github.com/travis-ci/travis-ci/issues/4756.

The cryptography project itself IS able to test >1.0 on the latest PyPy. They install pyenv in order to be able to pull in arbitrary versions of Python, and they use that to install the latest version of PyPy. https://github.com/pyca/cryptography/blob/master/.travis/install.sh#L57

This doesn't seem to negatively affect their ability to run with sudo: false. https://github.com/pyca/cryptography/blob/master/.travis.yml#L1

I might experiment with doing that, since I think that'd be better than testing with a pre-1.0 versions of cryptography. What do you think?

jmoldow:
Okay, I fixed the PyPy issue in master. Can you rebase or merge on master, and remove your PyPy / cryptography-related changes?

aptxkid:
Done. Thanks, @jmoldow !!

requirements.txt
```diff
@@ -1,4 +1,6 @@
 cryptography>=0.9.2
+enum34>=1.0.4
+ordereddict>=1.1
```
jmoldow: These need to be removed, they break the build on Python 3.5. I removed them in the commit I merged to master.

jmoldow: In case you're wondering, we still install them, but it happens in setup.py, protected by an if-statement for Python 2.

The dot at the bottom of this file makes sure that we install boxsdk itself, and also that we run setup.py. That will pick up enum34 and ordereddict when necessary.

BTW, can you change the ``. ``at the bottom of this file to ``-e .``? Thanks.

jmoldow:
Remove "I also restricted the version of cryptography to be <1.0. Otherwise it won't work for pypy." from your commit message.

boxsdk/auth/jwt_auth.py
```diff
@@ -44,6 +45,10 @@ def __init__(
             The ID of the Box Developer Edition enterprise.
         :type enterprise_id:
             `unicode`
+        :param jwt_key_id:
+            Key id for the jwt assertion
```
jmoldow: To be consistent, can you:
- change id to ID
- change jwt to JWT
- end with a period

test/unit/auth/test_jwt_auth.py
```diff
@@ -25,6 +25,11 @@ def jwt_algorithm(request):
     return request.param
 
 
+@pytest.fixture(params=('key_id_1', 'key_id_2'))
```
jmoldow: What benefits do we get from testing with two mostly-identical keys? Even if they were completely different, does this get us anything? In other test files, we don't e.g. test with two different object ids.

jmoldow:
:+1: but please address comments.

test/unit/auth/test_jwt_auth.py
aptxkid: No, I don't think it gives us much. I did this just to be consistent with other fixtures in this test module.

jmoldow:
This looks good now (though if you get a chance, maybe remove the part of your commit message that is no longer accurate).

Since this is a new required parameter, I think we should hold off on merging this until the public documentation is made available at https://box-content.readme.io/v2.0/docs/constructing-the-json-web-token-jwt-assertion.

This is a breaking change. But since Box Developer Edition is still in beta, we can probably do just a minor version bump.

aptxkid:
@jmoldow Sure. I wasn't planning merging it either. :)

Issue #71, opened 2015-09-16
Token Refresh in a Multiple Process Environment


walgitrus:
i’m finding an issue when making box sdk requests from a django app which runs with multiple worker processes. the scenario seems to be that one process makes a request which results in token auto-refresh. after this, any other worker process which makes a request will fail due to a now-expired token + expired refresh token.

the exception is
BoxOAuthException: 
Message: {"error":"invalid_grant","error_description":"Invalid refresh token"}
Status: 400
URL: https://api.box.com/oauth2/token
Method: POST

the stack trace is
boxsdk/session/box_session.py ? in _retry_request_if_necessary
boxsdk/session/box_session.py ? in _renew_session
boxsdk/auth/oauth2.py ? in refresh
boxsdk/auth/oauth2.py ? in _refresh
boxsdk/auth/oauth2.py ? in send_token_request

one possibility is to add a a pre-refresh callback, similar to the OAuth2._store_tokens callback.

Jeff-Meadows:
@walgitrus Please reopen if #81 doesn't solve your problem.

PR #72, opened 2015-09-16
make oauth2 token refresh more multi-process-friendly

walgitrus:
- add optional retrieve_tokens callback to OAuth2 to fetch possibly refreshed tokens from local store
- addresses issue #71

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -23,6 +23,7 @@ class OAuth2(object):
             client_id,
             client_secret,
             store_tokens=None,
+            retrieve_tokens=None,
             box_device_id='0',
             box_device_name='',
             access_token=None,
@@ -41,6 +42,10 @@ class OAuth2(object):
         :param store_tokens:
             Optional callback for getting access to tokens for storing them.
         :type store_tokens:
+            `callable`
+        :param retrieve_tokens:
+            Optional callback for retrieving tokens prior to refresh.
+        :type retrieve_tokens:
             `callable`
         :param box_device_id:
             Optional unique ID of this device. Used for applications that want to support device-pinning.
@@ -66,6 +71,7 @@ class OAuth2(object):
         self._client_id = client_id
         self._client_secret = client_secret
         self._store_tokens = store_tokens
+        self._retrieve_tokens = retrieve_tokens
         self._access_token = access_token
         self._refresh_token = refresh_token
         self._network_layer = network_layer if network_layer else DefaultNetwork()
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @walgitrus has signed the CLA. Thanks for the pull request!

walgitrus:
looks like `requirements.txt` should fix cryptography at 0.9.2 (or at least below 1.0) in order for pypy to work

jmoldow:
Thanks for the PR. First look through looks good. We'll fix the cryptography issue (there's a fix in #70, though I might try something else), and then look to merge this.

jmoldow:
I merged a fix for the PyPy issue. Going to close and re-open this PR to try to get Travis to generate the new test matrix.

walgitrus:
rebased to pull in the changes from master

boxsdk/auth/oauth2.py
```diff
@@ -28,6 +28,7 @@ def __init__(
             access_token=None,
             refresh_token=None,
             network_layer=None,
+            retrieve_tokens=None,
```
jmoldow: Since this is similar to store_tokens (and they'll probably be used together), let's put this below store_tokens in the argument list. Same in the docstring.

walgitrus: changes made and pushed

boxsdk/auth/oauth2.py
```diff
@@ -42,6 +43,10 @@ def __init__(
             Optional callback for getting access to tokens for storing them.
         :type store_tokens:
             `callable`
+        :param retrieve_tokens:
+            Optional callback for retrieving tokens prior to refresh.
```
jmoldow: Let's expand on this slightly.

We should specify that the callback takes no arguments, and that it should either return None (if no credentials are currently stored), or a (access token, refresh token) tuple.

And maybe we should add a brief description of what this feature would be used for. Something like "When using multiple processes or multiple separate OAuth2 objects, provide a retrieve_tokens function so that the same credentials can be reused across all your OAuth2 objects. The first object that needs to authenticate will do so, and store the tokens with the store_tokens callback (or the authentication and storage might happen entirely outside of the SDK). All subsequent objects, rather than re-authenticating, will use the tokens provided by retrieve_tokens. And the same behavior will occur when they all need to do token refresh." Feel free to reword or make that more concise.

jmoldow:
Hi @walgitrus,

We're going to handle this in PR #81 (see the `CooperativelyManagedOAuth2Mixin` class). If you want, take a look over there and see if the proposed interface will work for you.

PR #79, opened 2015-10-03
Rename uses of enterprise_token to enterprise_id

jmoldow:
This parameter is referred to as `enterprise_id` in the Box Developer documentation. Also, the noun "token" is already used to refer to the data returned at the end of the auth process. So rename this to reduce confusion.

This is a backwards compatible change, because the method parameter name was already correct. This is only a change to private attributes, test code, and documentation.

Also add a temporary pip version exclusion on the latest version of pytest, which is broken on Python 2. See https://github.com/pytest-dev/pytest/issues/1085 and https://github.com/pytest-dev/pytest/issues/1035.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -248,7 +248,7 @@ instead use an instance of `JWTAuth`.
     auth = JWTAuth(
         client_id='YOUR_CLIENT_ID',
         client_secret='YOUR_CLIENT_SECRET',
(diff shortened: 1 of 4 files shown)


boxcla:

potrebic:
Looks good to me +2

PR #81, opened 2015-10-14
Add network and auth subclasses.

Jeff-Meadows:
- Adds LoggingNetwork to allow clients to see what network calls the SDK is making.
- Adds RemoteOAuth2 to allow auth to occur in a different process or machine from the SDK.
- Adds RedisManagedOAuth2 to allow token storage in Redis, to enable multiple processes or machines to share tokens.
- Addresses issue #71 with `CooperativelyManagedOAuth2Mixin` class.

### DIFF
--- a/boxsdk/auth/cooperatively_managed_oauth2.py
+++ b/boxsdk/auth/cooperatively_managed_oauth2.py
@@ -1,0 +1,35 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+from boxsdk import OAuth2
+
+
+class CooperativelyManagedOAuth2Mixin(OAuth2):
+    """
+    Box SDK OAuth2 mixin.
+    Allows for sharing auth tokens between multiple clients.
+    """
+    def __init__(self, retrieve_tokens=None, *args, **kwargs):
+        """
+        :param retrieve_tokens:
+            Callback to get the current access/refresh token pair.
+        :type retrieve_tokens:
+            `callable` of () => (`unicode`, `unicode`)
+        """
+        self._retrieve_tokens = retrieve_tokens
+        super(CooperativelyManagedOAuth2Mixin, self).__init__(*args, **kwargs)
+
+    def _get_tokens(self):
+        """
+        Base class override. Get the tokens from the user-specified callback.
+        """
+        return self._retrieve_tokens()
+
+
+class CooperativelyManagedOAuth2(CooperativelyManagedOAuth2Mixin):
+    """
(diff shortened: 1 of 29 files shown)


boxcla:

Jeff-Meadows:
Obviously this needs tests, but wanted to review the ideas first and see if we should introduce these classes into the SDK.

boxsdk/auth/redis_managed_oauth2.py
```diff
@@ -0,0 +1,76 @@
...
+        self._update_current_tokens()
+        return super(RedisManagedOAuth2Mixin, self)._get_tokens()
+
+    def send_token_request(self, *args, **kwargs):
+        """
+        Base class override.
+        Saves the refreshed tokens in redis.
+        """
+        tokens = super(RedisManagedOAuth2Mixin, self).send_token_request(*args, **kwargs)
+        self._redis_server.hmset(self._unique_id, {'access': tokens[0], 'refresh': tokens[1]})
```
Jeff-Meadows: Whoops. I refactored the base class to make _store_tokens overridable to avoid overriding this, but then forgot to change it.

boxsdk/auth/cooperatively_managed_oauth2.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Box SDK OAuth2 mixin.
+    Allows for sharing auth tokens between multiple clients.
+    """
+    def __init__(self, retrieve_tokens=None, *args, **kwargs):
+        """
+        :param retrieve_tokens:
+            Callback to get the current access/refresh token pair.
+        :type retrieve_tokens:
+            `callable` of () => (`unicode`, `unicode`)
```
jmoldow: The callable might not be able to retrieve anything (e.g., the cache might not have this user yet). Should we allow the callable to return None, or allow the tuple elements to be None?

boxsdk/auth/cooperatively_managed_oauth2.py
```diff
@@ -0,0 +1,35 @@
...
+        super(CooperativelyManagedOAuth2Mixin, self).__init__(*args, **kwargs)
+
+    def _get_tokens(self):
+        """
+        Base class override. Get the tokens from the user-specified callback.
+        """
+        return self._retrieve_tokens()
+
+
+class CooperativelyManagedOAuth2(CooperativelyManagedOAuth2Mixin):
```
jmoldow: Should we create a concrete class for JWT as well?

PR #88, opened 2015-11-06
Add requests-toolbelt to setup.py

djgoku:
requests-toolbelt was added to the requirements.txt but not setup.py.

Is there a reason that both are needed or could it be consolidated to just using requirements.txt.


boxcla:
Verified that @djgoku has signed the CLA. Thanks for the pull request!

PR #90, opened 2015-11-16
Replace `basestring` with `six.string_types`

jmoldow:
On Python 3, calling `setup_logging()` in `boxsdk.util.log`
would always raise

    NameError: name 'basestring' is not defined

On Python 3, this prevented `LoggingNetwork` from being
initialized without passing an existing `Logger`.

Replace this usage of `basestring` with the equivalent
`six.string_types`.

Add unit test coverage of `boxsdk.util.log`.

Bump version to 1.3.2.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,16 @@ Release History
 Upcoming
 ++++++++
 
+1.3.2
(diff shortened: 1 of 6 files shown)


boxcla:

PR #107, opened 2016-01-08
Add membership() to User object

enmand:
Add paged `GroupMembership`s list to the Box User object

### DIFF
--- a/boxsdk/object/user.py
+++ b/boxsdk/object/user.py
@@ -1,11 +1,45 @@
 # coding: utf-8
 
 from __future__ import unicode_literals
+from functools import partial
 
 from .base_object import BaseObject
+from .group_membership import GroupMembership
 
 
 class User(BaseObject):
     """Represents a Box user."""
 
     _item_type = 'user'
+
+    def groups(self, starting_index=0, limit=100, include_page_info=False):
+        """
+        A generator over all of the Groups that this User is a member of.
+
+        :param starting_index:
+            The index at which to begin.
+        :type starting_index:
+            `int`
+        :param limit:
+            The maximum number of items to return in a page.
+        :type limit:
+            `int`
+        :returns:
+            A generator of GroupMembership instances. Or, if include_page_info
+            is True, it is a generator of 3-tuples, where each tuple is
+                1) GroupMembership instance
+                2) Number of GroupMemberships returned by the last paged API call
+                3) Index of *this* GroupMembership instance in the current page.
+        :rtype:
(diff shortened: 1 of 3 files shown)


boxcla:
Hi @enmand, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

enmand:
CLA signed

boxcla:
Verified that @enmand has just signed the CLA. Thanks, and we look forward to your contribution.

boxsdk/object/user.py
```diff
@@ -1,11 +1,45 @@
...
 from .base_object import BaseObject
+from .group_membership import GroupMembership
 
 
 class User(BaseObject):
     """Represents a Box user."""
 
     _item_type = 'user'
+
+    def groups(self, starting_index=0, limit=100, include_page_info=False):
```
jmoldow: This is called `groups`, but it returns `GroupMembership`s, not `Group`s.

Idea: rename this method to `membership` (then it mirrors `Group.membership()`). Optionally, add another method

```python
def groups(self, starting_index=0, limit=100, include_page_info=False):
    for group_membership, current_page_size, index_in_current_page in self.membership(starting_index=starting_index, limit=limit, include_page_info=True):
        group = group_membership.group
        if include_page_info:
            yield group, current_page_size, index_in_current_page
        else:
            yield group
```

boxsdk/object/user.py
```diff
@@ -1,11 +1,45 @@
...
+        """
+        A generator over all of the Groups that this User is a member of.
```
jmoldow: Lets make it clearer that these are `GroupMembership`s, not `Group`s.

jmoldow:
Looks good to me, except for the comments I'd like to see addressed. But @Jeff-Meadows  and @potrebic , can you also take a look at this?

Also, are you able to add tests? I'd like to see tests before we merge this.

enmand:
Sure, I'll add some tests.

Since `User.membership()` and `Group.membership()` are essentially the same, would it be better to have them inherit a common object which implements most of the `membership` func?

Jeff-Meadows:
Thanks for the PR.

* +1 for tests
* Can you add info about this change to HISTORY.rst?
* Optionally, add yourself to AUTHORS.rst

mattwiller:
We added [`user.get_group_memberships()`](https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.user.User.get_group_memberships) to the SDK, which supersedes this change

PR #114, opened 2016-02-10
Expose file.get_shared_link_download_url method.

### LINKED ISSUE #110: Add support for getting file shared link download URLs
The [get_shared_link()](https://github.com/box/box-python-sdk/blob/master/boxsdk/object/item.py#L162) method returns:

```python
return item.shared_link['url']
```

However sometimes there may be a use case for getting a **direct link** to a file, which should also be present in the response from the [Box API](https://box-content.readme.io/reference#create-a-shared-link-for-a-file):

```json
{
    "type": "file",
    "id": "10559150999",
   ...
    "shared_link": {
        "url": "https://foo.box.com/s/7mcmdlavtye5o5i0ue8xmtwh2sx5bv8p",
        "download_url":"https://foo.box.com/shared/static/7mcmdlavtye5o5i0ue8xmtwh2sx5bv8p.png",
    ...
    }
}
```

Currently the Box Python SDK has a common `get_shared_link` method on the `Item` class, and since `download_url` is not a valid attribute for a shared link on a folder this probably should be added *only* in the context of `File`.

I'm beginning to implement this on my fork with the intent to submit back upstream. What seems like the least terrible approach would be to clone the `get_shared_link` method from the `Item` class into the `File` class, rename it `get_download_link` (or `get_download_url`?), and change the return value, but this seems like a lot of code duplication. Is there a better way?

Any recommendations on approaches or proper patterns to stick to would be most appreciated!

**

[Description of one specific use case for this feature](https://community.box.com/t5/Admin-Forum/quot-We-re-sorry-but-we-can-t-preview-zip-files-quot-is/m-p/11336)

Jeff-Meadows:
Fixes #110

### DIFF
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -248,6 +248,7 @@ class File(Item):
     ):
         """
         Get a shared link download url for the file with the given access permissions.
+        This url is a direct download url for the file.
 
         :param access:
             Determines who can access the shared link. May be open, company, or collaborators. If no access is
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -237,3 +237,53 @@ class File(Item):
             :class:`Metadata`
         """
         return Metadata(self._session, self, scope, template)
+
+    def get_shared_link_download_url(
+            self,
+            access=None,
+            etag=None,
+            unshared_at=None,
+            allow_preview=None,
+            password=None,
+    ):
+        """
+        Get a shared link download url for the file with the given access permissions.
+
+        :param access:
+            Determines who can access the shared link. May be open, company, or collaborators. If no access is
+            specified, the default access will be used.
+        :type access:
+            `unicode` or None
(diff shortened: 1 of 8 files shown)


boxcla:

boxsdk/object/file.py
```diff
@@ -237,3 +237,53 @@ def metadata(self, scope='global', template='properties'):
...
+            self,
+            access=None,
+            etag=None,
+            unshared_at=None,
+            allow_preview=None,
+            password=None,
+    ):
+        """
+        Get a shared link download url for the file with the given access permissions.
+
```
aptxkid: Maybe a sentence or a link to explain the difference between shared link and direct url? I guess it might be confusing to a few users.

aptxkid:
:+1: a small comment on docstring.

nehalecky:
Hey wow! That was super fast, thanks all!

PR #115, opened 2016-02-15
Add get_shared_link_download_url() documentation/example to README

Frencil:
This is an addendum to #114 (which addressed #110) and merely adds documentation for the new `get_shared_link_download_url()` method to the top-level README, along with clarification for which types of items (files and/or folders) `get_shared_link()` and `get_shared_link_download_url()` can be used.


boxcla:
Verified that @Frencil has signed the CLA. Thanks for the pull request!

PR #121, opened 2016-03-07
Docs

Jeff-Meadows:


boxcla:

boxsdk/client/client.py
```diff
@@ -3,24 +3,34 @@
...
+from ..object.group import Group
+from ..object.group_membership import GroupMembership
+from ..util.shared_link import get_shared_link_header
+from ..util.translator import Translator
+from ..util.url import get_url
 
 
 class Client(object):
 
-    def __init__(self, oauth, network_layer=None, session=None):
```
jmoldow: I'd prefer to leave the oauth param required, than have ``Client.__init__`` slowly gain all of the params from ``OAuth.__init__``. And if you already have these params available, it really is barely more convenient to pass them straight to `Client`, as opposed to passing them to `OAuth2` which is passed to `Client`. I think we lose a lot more cluttering up the param list, than we gain in convenience.

boxsdk/client/client.py
```diff
@@ -34,12 +44,37 @@ def __init__(self, oauth, network_layer=None, session=None):
...
+            Ignored if oauth instance is passed into this constructor.
+        :type client_secret:
+            `unicode`
+        :param store_tokens:
+            Optional callback for getting access to tokens for storing them.
+            Ignored if oauth instance is passed into this constructor.
+        :type store_tokens:
+            `callable`
         """
         network_layer = network_layer or DefaultNetwork()
```
jmoldow: This can be ``network_layer = network_layer or  oauth.network_layer or DefaultNetwork()`` (would need to add a ``network_layer`` property to ``OAuth2``).

boxsdk/client/logging_client.py
```diff
@@ -0,0 +1,19 @@
...
+from ..network.logging_network import LoggingNetwork
+
+
+class LoggingClient(Client):
+    """
+    Box client subclass which logs requests and responses.
+    """
+    def __init__(self, oauth=None, network_layer=None, session=None, **kwargs):
+        super(LoggingClient, self).__init__(
+            oauth,
```
jmoldow: Might as well make this ``oauth=oauth``.

boxsdk/config.py
```diff
@@ -9,3 +9,9 @@ class API(object):
     UPLOAD_URL = 'https://upload.box.com/api/2.0'
     OAUTH2_API_URL = 'https://api.box.com/oauth2'  # <https://developers.box.com/docs/#oauth-2>
     OAUTH2_AUTHORIZE_URL = 'https://app.box.com/api/oauth2/authorize'  # <https://developers.box.com/docs/#oauth-2-authorize>
+
+
+class Version(object):
+    """Configuration object containing the package version and user agent string."""
+    VERSION = '1.5.0'
+    USER_AGENT_STRING = 'box-python-sdk-{0}'.format(VERSION)
```
jmoldow: Does this need to be a byte string? Or is unicode fine?

boxsdk/object/base_object.py
```diff
@@ -49,15 +49,17 @@ def __init__(self, session, object_id, response_object=None):
...
-        return self._response_object[item]
+        self.__dict__.update(self._response_object)
 
     def __getitem__(self, item):
         """Base class override. Try to get the attribute from the API response object."""
         return self._response_object[item]
 
+    def __repr__(self):
+        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
+        description = self.name if 'name' in self._response_object else self.object_id  # pylint:disable=no-member
```
jmoldow: We should always show the Box ID. And then also add the `name` if it is available.

We should also always show the `type`. ``self.__class__.__name__`` isn't always descriptive enough, if we haven't written a class / translator for that type.

jmoldow: Also, description could be its own property.

boxsdk/object/base_object.py
```diff
@@ -49,15 +49,17 @@ def __init__(self, session, object_id, response_object=None):
...
+        return '<Box {0} ({1})>'.format(self.__class__.__name__, description)
```
jmoldow: We need to ensure that the format string and description are native ``str`` objects. Python2 doesn't like it if repr returns unicode with non-ASCII characters.

boxsdk/session/box_session.py
```diff
@@ -81,7 +82,7 @@ def __init__(self, oauth, network_layer, default_headers=None):
         """
         self._oauth = oauth
         self._network_layer = network_layer
-        self._default_headers = default_headers or {}
+        self._default_headers = default_headers or {'User-Agent': Version.USER_AGENT_STRING}
```
jmoldow: Could this instead be

```python
self._default_headers = dict(default_headers)
self._default_headers.setdefault('User-Agent', Version.USER_AGENT_STRING)
self._default_headers = {header: header_value for header, header_value in six.iteritems(self._default_headers) if header_value is not None}
```

That way, we include the `User-Agent` even if they've defined their own headers, but we still allow them to turn off the `User-Agent` by passing `None`.

Also, does `'User-Agent'` need to be a byte string?

AUTHORS.rst
```diff
@@ -0,0 +1,9 @@
+The Box Python SDK is an open source project supported by `Box <https://box.com>`_
+used to interact with the Box API. This is a list of contributors.
+
+- `@Jeff-Meadows <https://github.com/Jeff-Meadows>`_
+- `@jmoldow <https://github.com/jmoldow>`_
+- `@aptxkid <https://github.com/aptxkid>`_
+- `@hnguyen08 <https://github.com/hnguyen08>`_
+- `@potrebic <https://github.com/potrebic>`_
+- `@nsundareswaran <https://github.com/nsundareswaran>`_
```
jmoldow: We've had more contributors than just this. https://github.com/box/box-python-sdk/graphs/contributors

Also, should we also allow full names? I have a vague preference for my name in addition to, or instead of, my Github handle (but also still include the link to my profile).

HISTORY.rst
```diff
@@ -6,6 +6,24 @@ Release History
 Upcoming
 ++++++++
 
+- Box objects have an improved ``__repr__``, making them easier to identify during debugging sessions.
+- Box objects now implement ``__dir__``, making them easier to explore. When created with a Box API response,
+  these objects will now include the API response fields as attributes.
```
jmoldow: Do you want to include these in the proposed 1.5.0 release?

README.rst
```diff
@@ -349,18 +347,47 @@ For advanced uses of the SDK, two additional auth classes are provided:
...
+
+Logging Client
+~~~~~~~~~~~~~~
 
-For more insight into the network calls the SDK is making, you can use the ``LoggingNetwork`` class. This class logs
+For more insight into the network calls the SDK is making, you can use the ``LoggingClient`` class. This class logs
 information about network requests and responses made to the Box API.
 
-.. code-block:: python
+.. code-block:: pycon
```
jmoldow: python

README.rst
```diff
@@ -349,18 +347,47 @@ For advanced uses of the SDK, two additional auth classes are provided:
...
+    >>> client.user().get()
+    GET https://api.box.com/2.0/users/me {'headers': {u'Authorization': u'Bearer ---------------------------kBjp',
+                 u'User-Agent': u'box-python-sdk-1.5.0'},
+     'params': None}
+    {"type":"user","id":"..","name":"Jeffrey Meadows","login":"..",..}
+    <boxsdk.object.user.User at 0x10615b8d0>
+
+For more control over how the information is logged, use the ``LoggingNetwork`` class directly.
+
+.. code-block:: pycon
```
jmoldow: python

README.rst
```diff
@@ -349,18 +347,47 @@ For advanced uses of the SDK, two additional auth classes are provided:
...
+~~~~~~~~~~~~~~~~~~
+
+For exploring the Box API, or to quickly get going using the SDK, the ``DevelopmentClient`` class combines the
+``LoggingClient`` with the ``DeveloperTokenClient``.
+
+Developer Token Client
+~~~~~~~~~~~~~~~~~~~~~~
+
+The Box Developer Console allows for the creation of short-lived developer tokens. The SDK makes it easy to use these
+tokens. Use the
```
jmoldow: Finish this thought.

boxsdk/__init__.py
```diff
@@ -3,5 +3,7 @@
 from __future__ import unicode_literals
```
jmoldow: absolute_import

boxsdk/client/__init__.py
```diff
@@ -0,0 +1,15 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+from .client import Client
+from .developer_token_client import DeveloperTokenClient
+from .logging_client import LoggingClient
+
+
+class DevelopmentClient(DeveloperTokenClient, LoggingClient):
```
jmoldow: Why is this in here, when all the others are in their own files?

boxsdk/client/client.py
```diff
@@ -3,24 +3,34 @@
 from __future__ import unicode_literals
```
jmoldow: absolute_import

boxsdk/__init__.py
```diff
@@ -3,5 +3,7 @@
...
 
 from .auth import JWTAuth, OAuth2
-from .client import Client, DeveloperTokenClient
+from .client import *  # pylint:disable=wildcard-import,redefined-builtin
 from .object import *  # pylint:disable=wildcard-import,redefined-builtin
+from .config import Version
+__version__ = Version.VERSION
```
jmoldow: Should there be newlines before this?

boxsdk/config.py
```diff
@@ -9,3 +9,9 @@ class API(object):
     UPLOAD_URL = 'https://upload.box.com/api/2.0'
     OAUTH2_API_URL = 'https://api.box.com/oauth2'  # <https://developers.box.com/docs/#oauth-2>
     OAUTH2_AUTHORIZE_URL = 'https://app.box.com/api/oauth2/authorize'  # <https://developers.box.com/docs/#oauth-2-authorize>
+
+
+class Version(object):
+    """Configuration object containing the package version and user agent string."""
+    VERSION = '1.5.0'
```
jmoldow: Now we have duplication between here and setup.py.

Can you do something like this:
- Create boxsdk/version.py, with contents
```python
__version__ = '1.5.0'
```
- In this file, ``from .version import __version__``, then ``VERSION = __version__``.
- In boxsdk/__init__.py, you can do ``from .version import __version__`` instead of ``__version__ = Version.VERSON``.
- In setup.py, ``open('boxsdk/version.py')`` and read the version constant. E.g. https://github.com/kennethreitz/requests/blob/master/setup.py#L50

boxsdk/config.py
```diff
@@ -9,3 +9,9 @@ class API(object):
     UPLOAD_URL = 'https://upload.box.com/api/2.0'
     OAUTH2_API_URL = 'https://api.box.com/oauth2'  # <https://developers.box.com/docs/#oauth-2>
     OAUTH2_AUTHORIZE_URL = 'https://app.box.com/api/oauth2/authorize'  # <https://developers.box.com/docs/#oauth-2-authorize>
+
+
+class Version(object):
```
jmoldow: Would prefer not to add more global variables. As it is, I have an in-progress change to try removing our reliance on the global ``API`` configs.

Can we instead get rid of ``Version.VERSION`` (it isn't a config and it isn't necessary; we can just define ``boxsdk.version.__version__`` as proposed below), and move ``USER_AGENT_STRING`` into a class attribute on ``BoxSession``?

boxsdk/util/url.py
```diff
@@ -0,0 +1,25 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+from boxsdk.config import API
+
+
+def get_url(endpoint, *args):
```
jmoldow: I have an in-progress change that changes how we use the ``API`` configs. Having this global function would not mesh well with what I had in mind.

Can we either make this private (so that we don't have to worry about future backwards-compatibility issues), or move it to be an instance method on ``BoxSession`` (where it can be used by ``BaseEndpoint`` and ``Client``)?

boxsdk/config.py
jmoldow: Oh, I didn't notice you already dealt with this.

setup.py
```diff
@@ -80,9 +81,11 @@ def main():
             python_conditional = 'python_version=="{0}"'.format(python_version)
             key = ':{0}'.format(python_conditional)
             extra_requires[key].append(requirement)
+    with open('boxsdk/config.py', 'r', 'utf-8') as config_py:
```
jmoldow: Use kwargs: ``encoding='utf-8'``.

boxsdk/client/client.py
Jeff-Meadows: I guess it's just a preference. One of the goals for the SDK is making it simple to get started. I've always disliked that you need to import OAuth2 and Client.

But I'm not married to this change. I can remove it for now and we can revisit.

boxsdk/config.py
Jeff-Meadows: This works in Py2 and Py3, so I don't think it needs to change.

AUTHORS.rst
Jeff-Meadows: I added those who are part of the Box org; additional contributors should feel free to add themselves, but I'd rather not add without their permission. Full names should be fine - anyone who wants that should make that change.

README.rst
Jeff-Meadows: This is python console, pycon

boxsdk/client/client.py
Jeff-Meadows: Not sure I want the client to infer its network layer from the oauth network layer

boxsdk/object/base_object.py
Jeff-Meadows: I'm just going to format and then convert to `str`. Good catch.

boxsdk/session/box_session.py
Jeff-Meadows: I think `requests` fixed the bytestring problem many releases ago, and we require a version later than the fixed version.

boxsdk/util/url.py
Jeff-Meadows: Sure, this can move to `Session`

boxsdk/config.py
Jeff-Meadows: I'd like `USER_AGENT_STRING` to be user configurable. However, I agree that version isn't a config.

version.py should work

PR #125, opened 2016-03-21
Add revoke() method to OAuth2 class.

### LINKED ISSUE #124: Add a way to revoke OAuth tokens
Using the OAuth2 class you can easily authenticate and refresh tokens but there is not an easy way to revoke the token.

Right now to revoke the token, we would have to do something like:
```python
data = {
  'client_id': CLIENT_ID,
  'client_secret': CLIENT_SECRET,
  'token': refresh_token
}
client.make_request(
  'POST',
  '{0}/revoke'.format(API.OAUTH2_API_URL),
  data=data
)
```

It would be nice if I could do something like `oauth.revoke()` to revoke and clear the access token and the refresh token.

Jeff-Meadows:
Fixes #124

The Box API includes a /revoke endpoint to de-auth tokens.
This commit makes it easy to use this endpoint with the SDK.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -275,7 +275,7 @@ class OAuth2(object):
                     'client_secret': self._client_secret,
                     'token': token_to_revoke,
                 },
-                access_token=token_to_revoke,
+                access_token=access_token,
             )
             if not network_response.ok:
                 raise BoxOAuthException(network_response.status_code, network_response.content, url, 'POST')
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -213,6 +213,8 @@ class OAuth2(object):
         return 'box_csrf_token_' + ''.join(ascii_alphabet[int(system_random.random() * ascii_len)] for _ in range(16))
 
     def _store_tokens(self, access_token, refresh_token):
+        self._access_token = access_token
+        self._refresh_token = refresh_token
         if self._store_tokens_callback is not None:
             self._store_tokens_callback(access_token, refresh_token)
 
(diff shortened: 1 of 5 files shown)


boxcla:

boxsdk/auth/oauth2.py
```diff
@@ -240,17 +242,40 @@ def send_token_request(self, data, access_token, expect_refresh_token=True):
...
+            url = '{base_auth_url}/revoke'.format(base_auth_url=API.OAUTH2_API_URL)
+            network_response = self._network_layer.request(
+                'POST',
+                url,
+                data={
+                    'client_id': self._client_id,
+                    'client_secret': self._client_secret,
+                    'token': token_to_revoke,
+                },
+                access_token=token_to_revoke,
```
jmoldow: What is this for?

boxsdk/auth/oauth2.py
```diff
@@ -240,17 +242,40 @@ def send_token_request(self, data, access_token, expect_refresh_token=True):
...
+        self._store_tokens(access_token, refresh_token)
         return self._access_token, self._refresh_token
+
+    def revoke(self):
+        """
+        Revoke the authorization for the current access/refresh token pair.
+        """
+        with self._refresh_lock:
+            token_to_revoke, _ = self._get_tokens()
+            if token_to_revoke is None:
```
jmoldow: If we did something like

```python
access_token, refresh_token = self._get_tokens()
token_to_revoke = access_token or refresh_token
```

Then we can support use cases where the user only stored the refresh token, not the access token, and now wants to revoke it.

boxsdk/auth/oauth2.py
```diff
@@ -240,17 +242,40 @@ def send_token_request(self, data, access_token, expect_refresh_token=True):
...
+                data={
+                    'client_id': self._client_id,
+                    'client_secret': self._client_secret,
+                    'token': token_to_revoke,
+                },
+                access_token=token_to_revoke,
+            )
+            if not network_response.ok:
+                raise BoxOAuthException(network_response.status_code, network_response.content, url, 'POST')
+            self._store_tokens(None, None)
```
jmoldow: Idea: add auto-revoking features:

```python
def close(self):
    return self.revoke()

def __enter__(self):
    return self

def __exit__(self, *exc_info):
    self.close()

def __del__(self):
    if self._revoke_tokens_on_del:
        self.close()
```

where `_revoke_tokens_on_del` is set from an ``__init__`` param `revoke_tokens_on_del` (defaulting to `False`).

Then, for applications that never want to leave active tokens around, they can be run like

```python
with OAuth2(*args, **kwargs) as oauth2:
    ...
```

or

```python
oauth2 = OAuth2(*args, revoke_tokens_on_del=True, **kwargs)
```

boxsdk/auth/oauth2.py
Jeff-Meadows: AFAIK, /revoke is an authenticated endpoint so we need to send an auth header. At any rate, Network.request always takes an `access_token` parameter, so we need to pass it in.

boxsdk/auth/oauth2.py
Jeff-Meadows: That seems nice.

boxsdk/auth/oauth2.py
Jeff-Meadows: Cool idea. Probably for a follow up PR

boxsdk/auth/oauth2.py
jmoldow: Where does this param end up being used? I can't find it in the code.

Jeff-Meadows: It's added to the `NetworkResponse`. This is necessary for our auto-refresh logic, to ensure we don't refresh the same token more than once.

jmoldow: Got it, thanks.

Should this always be `access_token`, instead of `token_to_revoke`?

PR #134, opened 2016-05-18
Always update `OAuth2` object with new tokens

### LINKED ISSUE #128: BoxSession._renew_session does not set previously refreshed tokens
I am using `CooperativelyManagedOAuth2` so that I can use the same oauth tokens for multiple clients but it seems like refreshed tokens do not get updated in the session if the tokens were refreshed by another client. I have two instances of the oauth class using the same tokens and two clients using one of the instances.

For example, say there are two clients initialized with the same tokens but different `CooperativelyManagedOAuth2` instances. The first client makes a request, sees the access token is expired and refreshes the tokens. The tokens in the first client's oauth instance are updated with the new tokens. The second client then makes a request, sees it's access token is expired and tries to refresh the token. The token is already refreshed so it does not try to refresh them again but then does not replace the old tokens so when client two retries the request it will raise a `BoxAPIException`.

jmoldow:
If the protected method `_get_tokens()` returns new tokens, refresh is skipped, and those tokens are used.

Previously, we would return the new tokens without updating the object state. Subsequent uses of the `OAuth2` object would use the old tokens.

Now we will correctly update the object state.

Fixes #128.

Also make some minor test fixes and upgrade the version of mock that we use.

Prepare for bump to version 1.5.2.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,11 @@ Release History
 Upcoming
 ++++++++
 
+1.5.2
+++++++++++++++++++
+
+- Bugfix so that ``OAuth2`` always has the correct tokens after a call to ``refresh()``.
+
 1.5.1 (2016-03-23)
 ++++++++++++++++++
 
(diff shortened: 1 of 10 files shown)


boxcla:

Jeff-Meadows:
👍

PR #139, opened 2016-06-16
Add event

### LINKED ISSUE #15: Add an Event class
Right now `get_events()` method in the `Events` class returns a `dict`. Ideally, we should have an `Event` class, and `Event` objects can also be translated. So `Events.get_events()` call will return a list of `Event` objects.

kelseymorris95:
Added an Event class, edited Events.get_event() to return a list of them.  Added basic Event unit test file for future use.

Fixes #15.


boxcla:
Hi @kelseymorris95, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

boxsdk/object/event.py
```diff
@@ -0,0 +1,21 @@
...
+class Event(BaseObject):
+    """Represents a single Box event"""
+
+    _item_type = 'event'
+
+    @property
+    def get_url(self, *args):
+        """
+        Base class override.
+        Disallow this method for this subclass, should not access Event by event_id.
```
HootyMcOwlface: Minor: Note that there is no get method in the box v2 api for Event.

test/unit/object/test_event.py
```diff
@@ -0,0 +1,23 @@
...
+
+from mock import Mock
+import pytest
+
+from boxsdk.object.event import Event
+
+
+def test_init_event(mock_box_session):
+    event = Event(mock_box_session, "f82c3ba03e41f7e8a7608363cc6c0390183c3f83", {
+       "type": "event",
```
HootyMcOwlface: formatting here... actually travis is probably complaining about this.
we usually use the format like:
event = Event(
  arg1,
  next_arg_on_new_line_if_we_are_using_multiline,
  {
    foo: bar,
    baz: wing,
    ding: {
      but_were_not_super_strict_I_didnt_put_the_above_curley_on_its_own_line: value1,
      but_we_are_picky_about_the_last_item_having_a_trailing_comma: value2,
    },
  },
)

test/unit/object/test_events.py
```diff
@@ -181,6 +182,18 @@ def mock_event():
 
 
 @pytest.fixture()
+def mock_event_object(mock_box_session):
```
HootyMcOwlface: consider renaming the mock_event fixture above to mock_event_json or something like that... just for clarity.

test/unit/object/test_events.py
```diff
@@ -181,6 +182,18 @@ def mock_event():
...
+    return Event(mock_box_session, "f82c3ba03e41f7e8a7608363cc6c0390183c3f83", {
```
HootyMcOwlface: This mock event object fixture should depend on the mock_event fixture (since it looks like it copies the json) and use that json blob to construct the event rather than copying.  This will make it easier to maintain the code; fewer places to change if the event json needs to be updated.

test/unit/object/test_events.py
```diff
@@ -205,6 +219,12 @@ def test_get_events(
         expected_url,
         params=dict(limit=100, stream_position=0, **expected_stream_type_params),
     )
+    event_entries = events['entries']
+    assert len(event_entries) == len(events_response.json.return_value['entries'])
+    if len(event_entries):
```
HootyMcOwlface: this could be a walk over the events in the list of event entries.  That way if the events_response gets modified to have more that 1 event returned then this test still works.

test/unit/object/test_events.py
```diff
@@ -253,7 +273,7 @@ def test_generate_events_with_long_polling(
         empty_events_response,
     ]
     events = test_events.generate_events_with_long_polling(**stream_type_kwargs)
-    assert next(events) == mock_event
+    assert next(events) == mock_event_object
```
HootyMcOwlface: instead of having the mock_event_object fixture you could construct the mock Event here based on the mock_event json.  Then do the assert check for equality.  Then I think you can remove the mock_event fixture.

boxsdk/object/event.py
```diff
@@ -0,0 +1,19 @@
...
+        """
+        raise AttributeError("'Event' class has no method 'get_url'")
```
jmoldow: Glad that you noticed this issue.

To me, this sort of thing is a code smell. If a baseclass defines a default behavior, and then a subclass is incapable of performing that behavior, chances are, the subclass should not actually inherit from that baseclass.

I believe that to be the case here. `BaseObject` is a class that is the root for all possible persistent, addressable resource types in the REST API.

Since we've never thought about non-persistent / non-addressable objects before, we didn't think about this when we first created #15.

I have an idea of what we should do here, but I'm curious for your thoughts.

Jeff-Meadows: I was thinking the same thing - `Event` should probably not derive directly from `BaseObject`.

@kelseymorris95 @jmoldow should we get together to discuss the right plan for this?

kelseymorris95: Yeah, I had discussed this with Jeremy because I wasn't sure what approach was best.  Most methods in `BaseObject` use get_url(), so I think the only behaviours of `BaseObject` that we actually want to mimic are __init__, __getitem__, __repr__, object_id, and __eq__.  
Is there an option besides reproducing those behaviors separately in `Event`?

jmoldow: Check out #140. The first half is unrelated, but the second half overlaps with what we're talking about here.

kelseymorris95:
CLA signed

boxcla:
Verified that @kelseymorris95 has just signed the CLA. Thanks, and we look forward to your contribution.

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,12 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
```
jmoldow: We should also be using `absolute_import` in all files.

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,12 @@
...
+from collections import Mapping
+
+from boxsdk.object.base_api_json_object import BaseAPIJSONObject
```
jmoldow: With `absolute_import` turned on, this should be changed to `from .base_api_json_object import BaseAPIJSONObject`.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,64 @@
...
+
+from __future__ import unicode_literals
+from abc import ABCMeta
+
+import six
+
+from boxsdk.util.translator import Translator
+
+
+class BaseAPIJSONObjectMeta(ABCMeta):
```
jmoldow: `ABCMeta` is for abstract baseclasses.

For metaclasses, just use `type`. I don't know why it was being used for `ObjectMeta` (probably an accident), but it wasn't necessary.

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class APIJSONObject(BaseAPIJSONObject, Mapping):
+    """Class representing objects that are not part of the REST API"""
```
jmoldow: Nitpick: end sentences with a period (here and in other places)

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,64 @@
...
+        item_type = attrs.get('_item_type', None)
+        if item_type is not None:
+            Translator().register(item_type, cls)
+
+
+@six.add_metaclass(BaseAPIJSONObjectMeta)
+class BaseAPIJSONObject(object):
+    """Base class containing basic logic shared between true REST objects and other objects (such as an Event)"""
+
+    def __init__(self, object_id=None, response_object=None):
```
jmoldow: Just like you got rid of `session` here, I think we can also get rid of `object_id`. Non-REST objects won't tend to have an `object_id`, so I think it'd be fine to omit it. That functionality can be directly on `BaseObject`, which is the only subclass that needs it.

Don't forget a docstring for this.

And since we can't guarantee the existence of an `object_id`, I think that we probably do want to require the existence of the `response_object` (I don't think we'll be creating `BaseAPIJSONObject` objects without them, unless they are `BaseObject`s). So I think we can make the signature be

```python
def __init__(self, response_object):
```

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,64 @@
...
+        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
+        description = '<Box {0} - {1}>'.format(self.__class__.__name__, self._description)
+        if six.PY2:
+            return description.encode('utf-8')
+        else:
+            return description
+
+    @property
+    def _description(self):
+        if 'name' in self._response_object:
```
jmoldow: I can't think of any non-REST objects which will have a `name`, and most won't have an `object_id`. So most non-REST objects won't have any description at all. I think the default return value of `_description` can be the empty string. This particular implementation can stay on `BaseObject`.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,64 @@
...
+        self._response_object = response_object or {}
+        self.__dict__.update(self._response_object)
+
+    def __getitem__(self, item):
+        """Base class override. Try to get the attribute from the API response object."""
+        return self._response_object[item]
+
+    def __repr__(self):
+        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
+        description = '<Box {0} - {1}>'.format(self.__class__.__name__, self._description)
```
jmoldow: (Read my next comment first.)

Since some objects might not have any description, we might want to change this to be something like

```python
extra_description = ' - {0}'.format(self._description) if self._description else ''
description = '<Box {0}{1}>'.format(self.__class__.__name__, extra_description)
```

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,64 @@
...
+        """Return the Box ID for the object.
+
+        :rtype:
+            `unicode`
+        """
+        return self._object_id
+
+    def __eq__(self, other):
+        """Base class override. Equality is determined by object id."""
+        return self._object_id == other.object_id
```
jmoldow: Without `object_id`, equality can either be:
- By reference only (in which case, we can get rid of this method).
- By equality of `self._response_object`.

boxsdk/object/base_endpoint.py
```diff
@@ -2,11 +2,13 @@
 
 from __future__ import unicode_literals
 
+from boxsdk.object.base_api_json_object import BaseAPIJSONObject
 
-class BaseEndpoint(object):
+
+class BaseEndpoint(BaseAPIJSONObject):
```
jmoldow: Endpoints are more general than objects. For example, see https://github.com/box/box-python-sdk/blob/master/boxsdk/object/search.py#L120

So `BaseEndpoint` should remain unchanged. And `BaseObject` should inherit from both `BaseEndpoint` and `BaseAPIJSONObject`.

boxsdk/object/event.py
```diff
@@ -0,0 +1,11 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+from boxsdk.object.base_api_json_object import BaseAPIJSONObject
+
+
+class Event(BaseAPIJSONObject):
```
jmoldow: We want this to inherit from `APIJSONObject`, not the base.

boxsdk/object/base_api_json_object.py
kelseymorris95: In `Client`, there are methods `user`, `group`, `file`, and `folder` which return instances of their respective classes.  These are initialized with the `session` and the `object_id` of the object we want to use, but no `response_object` because we have not made a request.  If I make `response_object` a required field, should I initialize these objects with an empty dictionary?

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,27 @@
...
+
+from __future__ import unicode_literals, absolute_import
+from collections import Mapping
+
+# import six
+
+from .base_api_json_object import BaseAPIJSONObject, BaseAPIJSONObjectMeta
+
+
+class APIJSONObjectMeta(BaseAPIJSONObjectMeta, Mapping.__metaclass__):
```
jmoldow: Rather than using `Mapping.__metaclass__` (which won't work on Python 3), I would instead recommend either `Mapping.__class__`, `type(Mapping)`, or, since we already know what its metaclass is, `ABCMeta`.

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,27 @@
...
+    """Avoid conflicting metaclass definitions for APIJSONObject"""
+    pass
+
+
+# @six.add_metaclass(APIJSONObjectMeta)
+# class APIJSONObject(APIJSONObjectMeta, BaseAPIJSONObject, Mapping):
+class APIJSONObject(BaseAPIJSONObject, Mapping):
+    """Class representing objects that are not part of the REST API."""
+
+    __metaclass__ = APIJSONObjectMeta
```
jmoldow: This won't work on Python 3. Instead, you should do:

```python
class APIJSONObject(six.with_metaclass(APIJSONObjectMeta, BaseAPIJSONObject, Mapping)):
    ...
```

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __len__(self):
+        return len(self._response_object)
+
+    def __iter__(self):
+        return self._response_object.__iter__()
```
jmoldow: `return iter(self._response_object)` would be more Pythonic.

boxsdk/object/base_api_json_object.py
```diff
@@ -37,28 +41,13 @@ def __getitem__(self, item):
 
     def __repr__(self):
         """Base class override. Return a human-readable representation using the Box ID or name of the object."""
-        description = '<Box {0} - {1}>'.format(self.__class__.__name__, self._description)
+        extra_description = ' - {0}'.format(self._description) if self._description else ''
+        description = '<Box {0}{1}'.format(self.__class__.__name__, extra_description)
```
jmoldow: Missing > at the end of that string.

boxsdk/object/base_api_json_object.py
```diff
@@ -37,28 +41,13 @@ def __getitem__(self, item):
...
         if six.PY2:
             return description.encode('utf-8')
         else:
             return description
 
     @property
     def _description(self):
```
jmoldow: Since this is now a protected (instead of private) method that we'll be overriding in subclasses, we should add a docstring.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,53 @@
...
+        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
+        extra_description = ' - {0}'.format(self._description) if self._description else ''
+        description = '<Box {0}{1}'.format(self.__class__.__name__, extra_description)
+        if six.PY2:
+            return description.encode('utf-8')
+        else:
+            return description
+
+    @property
+    def _description(self):
```
jmoldow: Since this is now a protected (instead of private) method that we'll be overriding in subclasses, we should add a docstring.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,53 @@
...
+        self.__dict__.update(self._response_object)
+
+    def __getitem__(self, item):
+        """Base class override. Try to get the attribute from the API response object."""
+        return self._response_object[item]
+
+    def __repr__(self):
+        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
+        extra_description = ' - {0}'.format(self._description) if self._description else ''
+        description = '<Box {0}{1}'.format(self.__class__.__name__, extra_description)
```
jmoldow: Missing > at the end of that string.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,53 @@
...
+@six.add_metaclass(BaseAPIJSONObjectMeta)
+class BaseAPIJSONObject(object):
+    """Base class containing basic logic shared between true REST objects and other objects (such as an Event)"""
+
+    def __init__(self, response_object=None):
+        """
+        :param response_object:
+            The Box API response representing the object.
+        :type response_object:
+            :class:`BoxResponse`
```
jmoldow: This is just a plain `dict`. `BoxResponse` is something else.

Looks like this was a typo in the original documentation. Can you fix it?

While we're at it, maybe we can give this a better description as well. Perhaps "A JSON object, from a Box API response, representing the object." or something like that.

boxsdk/object/base_endpoint.py
```diff
@@ -1,18 +1,21 @@
...
 
-    def __init__(self, session):
+    def __init__(self, session, response_object=None):
         """
-
         :param session:
             The Box session used to make requests.
         :type session:
             :class:`BoxSession`
+        :param response_object:
```
jmoldow: This has no meaning or use here. `BaseEndpoint` should remain as it was.

boxsdk/object/base_object.py
```diff
@@ -1,34 +1,18 @@
 # coding: utf-8
 
-from __future__ import unicode_literals
-from abc import ABCMeta
+from __future__ import unicode_literals, absolute_import
 import json
 
-import six
+from .base_endpoint import BaseEndpoint
+from.base_api_json_object import BaseAPIJSONObject
```
jmoldow: A space after "from"

boxsdk/object/base_endpoint.py
```diff
@@ -1,18 +1,21 @@
...
+            The Box API response representing the object.
+        :type response_object:
+            :class:`BoxResponse`
         """
         self._session = session
```
jmoldow: Though, I would recommend adding back

```python
super(BaseEndpoint, self).__init__(**kwargs)
```

since we're now dealing with multiple-inheritance.

Add ``**kwargs`` to the method declaration. You might possibly that in some of the other ``__init__`` methods too.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,53 @@
...
+        item_type = attrs.get('_item_type', None)
+        if item_type is not None:
+            Translator().register(item_type, cls)
+
+
+@six.add_metaclass(BaseAPIJSONObjectMeta)
+class BaseAPIJSONObject(object):
+    """Base class containing basic logic shared between true REST objects and other objects (such as an Event)"""
+
+    def __init__(self, response_object=None):
```
jmoldow: Since we're going to be using multiple-inheritance, you should also give this method a `**kwargs`.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        :param response_object:
+            The Box API response representing the object.
+        :type response_object:
+            :class:`BoxResponse`
+        """
```
jmoldow: We need a `super()` here:

```python
super(BaseAPIJSONObject, self).__init__(**kwargs)
```

boxsdk/object/base_object.py
```diff
@@ -1,34 +1,18 @@
...
-            Translator().register(item_type, cls)
-
-
-@six.add_metaclass(ObjectMeta)
-class BaseObject(BaseEndpoint):
-    """
-    A Box API endpoint for interacting with a Box object.
-    """
+    # Question:
+    # Should this item type be a part of other objects?
```
jmoldow: No, it isn't necessary. This is just here so that all of its subclasses at least have `_item_type` defined, even if it is `None`.

Though this should move to `BaseAPIJSONObject`, since that is now the base for translation.

boxsdk/object/base_object.py
```diff
@@ -46,22 +30,14 @@ def __init__(self, session, object_id, response_object=None):
         :type response_object:
             :class:`BoxResponse`
         """
-        super(BaseObject, self).__init__(session)
```
jmoldow: ``super(BaseObject, self).__init__(session=session, response_object=response_object)`` should work, with the other `__init__` changes I mentioned. You should remove the manual `__init__` calls below.

boxsdk/object/base_object.py
```diff
@@ -28,7 +30,30 @@ def __init__(self, session, object_id, response_object=None):
...
+        # super(BaseObject, self).__init__(session, response_object)
+        BaseEndpoint.__init__(self, session, response_object)
+        BaseAPIJSONObject.__init__(self, response_object)
+
+    def __eq__(self, other):
+        """Equality is determined by object id"""
+        return self._object_id == other.object_id
+
+    @property
+    def _description(self):
```
jmoldow: "Baseclass override" docstring.

boxsdk/object/base_object.py
```diff
@@ -46,22 +30,14 @@ def __init__(self, session, object_id, response_object=None):
...
+        BaseAPIJSONObject.__init__(self, response_object)
 
-    def __repr__(self):
-        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
-        description = '<Box {0} - {1}>'.format(self.__class__.__name__, self._description)
-        if six.PY2:
-            return description.encode('utf-8')
-        else:
-            return description
+    def __eq__(self, other):
```
jmoldow: If you move this back to the position it used to be in, then it won't appear in the diff, simplifying things. Same with `get_url` and `get_type_url`.

jmoldow:
Something to think about: `Translator.translate()` takes a type, and tries to translate it. If it fails, it'll return `BaseObject`. Then, the usual procedure is to pass `session, object_id, response_object` to the class that is returned.

This has two problems:
- Ideally, the default return value would be either `BaseObject` or `APIJSONObject`, depending on the object. I would choose `BaseObject` if there were an `id`, and `APIJSONObject` otherwise. However, we can't do that since the method only accepts type. We'd need to make a backwards-incompatible change to make it accept additional information.
- The types of classes we return have different `__init__` parameters. So you can't initialize an object unless you know in advance which type it's going to be. We could fix this by:
  - Moving the object creation inside of `Translator.translate()` (after dealing with the above problem).
  - Making `APIJSONObject.__init__` have the same argument list as `BaseObject.__init__`, even though it doesn't need most of them.

Let's not worry about this right now. What we have now works well enough to push what you have as-is. But we'll want to revisit this before or during work on #140.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+import six
+
+from ..util.translator import Translator
+
+
+class BaseAPIJSONObjectMeta(type):
```
HootyMcOwlface: why inherit from type here instead of ABCMeta?

boxsdk/object/api_json_object.py
```diff
@@ -0,0 +1,23 @@
...
+
+from __future__ import unicode_literals, absolute_import
+from collections import Mapping
+from abc import ABCMeta
+import six
+
+from .base_api_json_object import BaseAPIJSONObject, BaseAPIJSONObjectMeta
+
+
+class APIJSONObjectMeta(BaseAPIJSONObjectMeta, ABCMeta):
```
HootyMcOwlface: This is an empty ABCMeta which is a little strange to me.  I don't follow the comment about avoiding conflicts.  We get BetaAPIJSONObjectMeta thru BaseAPIJSONObject below in APIJSONObject... so, I'm not sure what this is needed for.

HootyMcOwlface: Thnx for explaining... I think that's worth a link & comment.
http://code.activestate.com/recipes/204197-solving-the-metaclass-conflict/
or something saying this addresses the known issue with inheriting from 2 classes with separate metaclasses.

HootyMcOwlface:
:+1:

boxsdk/object/base_api_json_object.py
```diff
@@ -25,29 +24,45 @@ def __init__(cls, name, bases, attrs):
...
+    _item_type = None
+
+    def __init__(self, response_object=None, **kwargs):
         """
         :param response_object:
-            The Box API response representing the object.
+            A JSON object representing the object returned from a Box API request.
         :type response_object:
-            :class:`BoxResponse`
+            :`dict`
```
jmoldow: No colon here and in other places.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+    """Base class containing basic logic shared between true REST objects and other objects (such as an Event)"""
+
+    _item_type = None
+
+    def __init__(self, response_object=None, **kwargs):
+        """
+        :param response_object:
+            A JSON object representing the object returned from a Box API request.
+        :type response_object:
+            :`dict`
```
jmoldow: No colon here and in other places.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+        :param kwargs:
+            Keyword arguments for base class constructors.
+        :type kwargs:
+            :`dict`
+        """
+        self._response_object = response_object or {}
+        self.__dict__.update(self._response_object)
+        super(BaseAPIJSONObject, self).__init__(**kwargs)
```
jmoldow: Except when necessary, `super()` should happen first.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __getitem__(self, item):
+        """
+        Try to get the attribute from the API response object.
```
jmoldow: You need a blank line after the title line. Otherwise, sphinx won't be able to parse this.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+        :param item:
+            The attribute to retrieve from the API response object.
+        :type item:
+            `unicode`
+        :rtype:
+            `unicode`
```
jmoldow: Actually, the return value can be an arbitrary object:
- `unicode` if the field is a string.
- `datetime` if the field is a timestamp.
- Or `bool`, or `NoneType`, or `int`.
- Can also be a JSON `dict` or `list`.
- In the future, when we implement #140, it will be a `BaseAPIJSONObject`.

So I would either document this as returning `object`, or just not document anything. The reason I say you can get away with not documenting it, is because most magic methods have well-defined signatures. In the case of `__getitem__` for a generic `Mapping`, we know it takes a `Hashable`, and returns some `object`. Since we're not returning a more specific type here, we don't necessarily need to document it.

boxsdk/object/base_api_json_object.py
```diff
@@ -25,29 +24,45 @@ def __init__(cls, name, bases, attrs):
...
-        description = '<Box {0}{1}'.format(self.__class__.__name__, extra_description)
+        description = '<Box {0}{1}>'.format(self.__class__.__name__, extra_description)
         if six.PY2:
             return description.encode('utf-8')
         else:
             return description
 
     @property
     def _description(self):
+        """Return a description if the object if one exists."""
```
jmoldow: "of the object"

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+        extra_description = ' - {0}'.format(self._description) if self._description else ''
+        description = '<Box {0}{1}>'.format(self.__class__.__name__, extra_description)
+        if six.PY2:
+            return description.encode('utf-8')
+        else:
+            return description
+
+    @property
+    def _description(self):
+        """Return a description if the object if one exists."""
```
jmoldow: "of the object"

boxsdk/object/base_endpoint.py
```diff
@@ -1,20 +1,24 @@
...
             The Box session used to make requests.
         :type session:
             :class:`BoxSession`
+        :param kwargs:
+            Keyword arguments for base class constructors.
+        :type kwargs:
+            :`dict`
         """
         self._session = session
+        super(BaseEndpoint, self).__init__(**kwargs)
```
jmoldow: `super` should be first

boxsdk/object/base_endpoint.py
```diff
@@ -1,20 +1,24 @@
...
         """
-
         :param session:
             The Box session used to make requests.
         :type session:
             :class:`BoxSession`
+        :param kwargs:
+            Keyword arguments for base class constructors.
+        :type kwargs:
+            :`dict`
```
jmoldow: No colon here.

boxsdk/object/base_api_json_object.py
```diff
@@ -0,0 +1,68 @@
...
+
+    _item_type = None
+
+    def __init__(self, response_object=None, **kwargs):
+        """
+        :param response_object:
+            A JSON object representing the object returned from a Box API request.
+        :type response_object:
+            :`dict`
+        :param kwargs:
```
jmoldow: Typically don't need to document `**kwargs` when its only intent is to pass it to `super()`. This usage of `**kwargs` is well-understood. Same for similar documentation on other classes.

It's only when you're doing something else with `*args` or `**kwargs` that it would need to be documented.

boxsdk/object/base_object.py
```diff
@@ -42,29 +22,16 @@ def __init__(self, session, object_id, response_object=None):
...
-        return self._response_object[item]
-
-    def __repr__(self):
-        """Base class override. Return a human-readable representation using the Box ID or name of the object."""
-        description = '<Box {0} - {1}>'.format(self.__class__.__name__, self._description)
-        if six.PY2:
-            return description.encode('utf-8')
-        else:
-            return description
+        super(BaseObject, self).__init__(session=session, response_object=response_object)
```
jmoldow: `super` should be first.

boxsdk/object/item.py
```diff
@@ -111,6 +110,10 @@ def rename(self, name):
     def get(self, fields=None, etag=None):
         """Base class override.
 
+        :param fields:
```
jmoldow: This is fine to leave as-is. Though the reason it wasn't documented before is that we usually don't re-document params for baseclass overrides.

test/unit/object/test_api_json_object.py
```diff
@@ -0,0 +1,3 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
```
jmoldow: I would add a simple test in here. Just verify that if you instantiate this with a `dict`, that it has the same length, and is equal to the original dict.

test/unit/object/test_base_api_json_object.py
```diff
@@ -0,0 +1,18 @@
...
+
+from __future__ import unicode_literals, absolute_import
+import pytest
+
+from boxsdk.object.base_api_json_object import BaseAPIJSONObject
+
+
+@pytest.fixture(params=[{'foo': 'bar'}, {'a': {'b': 'c'}}])
+def base_api_json_object(request):
+    return request.param, BaseAPIJSONObject(request.param)
```
jmoldow: These could be two separate fixtures. One for the vanilla `dict`, and then another (depending on the first) which is the `BaseAPIJSONObject`. And then use both of those fixtures in the test method.

test/unit/object/test_events.py
```diff
@@ -205,6 +206,10 @@ def test_get_events(
         expected_url,
         params=dict(limit=100, stream_position=0, **expected_stream_type_params),
     )
+    event_entries = events['entries']
+    assert len(event_entries) == len(events_response.json.return_value['entries'])
```
jmoldow: I believe comparing them directly for equality (without `len`) would also work.

jmoldow:
👍 

Thanks @kelseymorris95 ! This all looks great. I'm going to merge it now.

Before releasing this to PyPI, over the next few days, let's make sure we're happy with the names of the new classes, make a final decision about `Mapping` vs `MutableMapping` vs `dict`, and double-check that we aren't introducing any other backwards incompatibilities.

PR #149, opened 2016-07-08
Allow custom subclasses of smart objects

kelseymorris95:
Enabled `Translator` to register new subclasses of smart objects.  Removed all hard coded smart object return types in favor of a lookup in `Translator` or returning an instance of `self.__class__`.


boxcla:

jmoldow:
👍 awesome!

PR #151, opened 2016-07-14
Changed base_object so it can be compared to None object

jhogg:
Changed __eq__ comparison to allow an object to be compared to None in addition to another object by id.   The current code requires that it be compared against another instantiated object.

### DIFF
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -153,7 +153,7 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
 
     def __eq__(self, other):
         """Equality as determined by object id"""
-        return self._object_id == other.object_id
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @jhogg, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

jhogg:
CLA Signed

boxcla:
Verified that @jhogg has just signed the CLA. Thanks, and we look forward to your contribution.

jmoldow:
Thanks for the pull request.

However, it would be more flexible to do something like

```python
if not isinstance(other, BaseObject):
    return NotImplemented
...
```

There are other bugs with `__eq__` as well, which I've documented in #163. I'll be opening up a new PR with these fixes.

PR #154, opened 2016-08-01
Cloneable

kelseymorris95:
Includes proof of concept from https://github.com/Jeff-Meadows/box-python-sdk/tree/api_wrapper.  New api_call decorator can be used to allow methods that make an API call to accept an extra parameter, to be used for sending low level request parameters to Requests or whatever network layer is being used.


boxcla:

boxsdk/client/client.py
```diff
@@ -13,11 +13,13 @@
 from ..object.file import File
 from ..object.group import Group
 from ..object.group_membership import GroupMembership
+from ..object.cloneable import Cloneable
```
Jeff-Meadows: alphabetize

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,54 @@
...
+    :rtype:
+        :class:`APICallWrapper`
+    """
+    wrapper = APICallWrapper(func)
+    update_wrapper(wrapper, func)
+    return wrapper
+
+
+class APICallWrapper(object):
+    _NOT_PRESENT = object()
```
Jeff-Meadows: I think this is not needed

Jeff-Meadows:
👍

boxsdk/object/cloneable.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import unicode_literals, absolute_import
+
+
+class Cloneable(object):
+    """
+    Cloneable interface to be implemented by endpoint objects that should have ability to be cloned, but with a
+    different session member if desired.
+    """
+
+    def __init__(self, *args, **kwargs):
```
jmoldow: This override isn't necessary, since all it does is call `super`.

boxsdk/object/cloneable.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def as_user(self, user):
+        """
+        Returns a new endpoint object with default headers set up to make requests as the specified user.
+        :param user:
+            The user to impersonate when making API requests.
+        :type user:
+            :class:`User`
+        """
+        raise NotImplementedError
```
jmoldow: `as_user` and `with_shared_link` will only ever be defined one way:

```python
def as_user(self, user):
    return self.clone(self.session.as_user(user))

def with_shared_link(self, shared_link, shared_link_password=None):
    return self.clone(self.session.with_shared_link(shared_link, shared_link_password=shared_link_password))
```

So I think we could move those definitions to here, and then remove that code from other classes.

Doing that would require adding `session` as a `property` or `abstractproperty` of `Cloneable`.

Alternatively, we could do

```python
def __init__(self, session, **kwargs):
    super(Cloneable, self).__init__(**kwargs)
    self._session = session
```

and have `Client` and `BaseEndpoint` pass `session` to `super` instead of setting `self._session` themselves. And then those method definitions I wrote above, can use `self._session` instead of `self.session`.

boxsdk/client/client.py
```diff
@@ -6,18 +6,20 @@
...
 from ..network.default_network import DefaultNetwork
-from ..object.user import User
+from ..object.cloneable import Cloneable
 from ..object.folder import Folder
 from ..object.search import Search
 from ..object.events import Events
 from ..object.file import File
 from ..object.group import Group
 from ..object.group_membership import GroupMembership
+from ..object.user import User
```
jmoldow: You should merge master. Not only is your PR not mergeable right now, but I also think that this import is no longer used on master.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+    def __get__(self, _instance, owner):
+        def call(*args, **kwargs):
+            instance = _instance
+            if instance is None:
+                # If this is being called as an unbound method, the instance is the first arg.
+                if owner is not None and len(args) > 0 and isinstance(args[0], owner):
+                    instance = args[0]
+                    args = args[1:]
+                else:
+                    raise AttributeError
```
jmoldow: I think it makes more sense for this to be `TypeError`.

`AttributeError` would imply that when doing `foo.bar`, the attribute `bar` couldn't be found. But what is happening here is that `foo.bar` was found fine, but `foo.bar(...)` had an error.

Also, `TypeError` matches what happens for Python builtins and stdlib:

```
In [1]: list.append()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-9190ca1dd107> in <module>()
----> 1 list.append()

TypeError: descriptor 'append' of 'list' object needs an argument

In [2]: list.append(5)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-bc24911c8611> in <module>()
----> 1 list.append(5)

TypeError: descriptor 'append' requires a 'list' object but received a 'int'

File: boxsdk/util/api_call_decorator.py [hunk 7]
```diff
@@ -0,0 +1,53 @@
...
+            request_data = kwargs.pop('request_data', {})
+            if request_data:
+                # If request_data is specified, then clone the instance, and specify it as the default network args.
+                # pylint: disable=protected-access
+                instance = instance.clone(instance._session.with_default_network_request_kwargs(**request_data))
```
jmoldow: This will slow down API calls, since every time you make the call, you'll need to spend time doing this cloning.

That's fine, especially since this feature probably won't be used by the average app. 

We don't necessarily need this for V1 (since it'll only be us using it immediately), but other things I'd be interested in seeing in the future:

- The ability to pass request_data to the decorator. So someone could do

```python
class MyFile(File):
    @api_call(request_data={'foo': 'bar'})
    def get(self, *args, **kwargs):
        ...
```
and then have that clone only happen once per object per method. This would only help if the user happens to be re-using objects for multiple API calls of the same type, which might not be common (so it might never be worth doing this).

- The ability to specify default `request_data` for all methods on an object. So someone could do

```python
class MyFile(File):
    default_request_data = {'foo': 'bar'}
```

or 
```python
class MyFile(File):
    def __init__(self, ..., foo):
        self.default_request_data = {'foo': foo}
```

and have the cloning happen at init time. Then hopefully some method calls might never need `request_data` passed to them at call time.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+        def call(*args, **kwargs):
+            instance = _instance
+            if instance is None:
+                # If this is being called as an unbound method, the instance is the first arg.
+                if owner is not None and len(args) > 0 and isinstance(args[0], owner):
+                    instance = args[0]
+                    args = args[1:]
+                else:
+                    raise AttributeError
+            request_data = kwargs.pop('request_data', {})
```
jmoldow: Did you want to keep this name, or use something else? Chris had suggested `network_parameters`.

Some other ideas: `network_arguments`, `extra_network_arguments`, `extra_request_arguments`, `extra_network_parameters`, `custom_network_parameters`, `custom_network_arguments`, `custom_request_arguments`, `custom_request_parameters`.

Also, in a future version, it might be cool to allow header customization in the same way. But we don't need that right now.

jmoldow: Also, since we won't use `request_data` if it is empty, we can just do `kwargs.pop('request_data', None)`. That way we don't need to construct a new `dict` on every API call.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+
+def api_call(func):
+    """
+    Designates the decorated method as one that makes a Box API call.
+    The decorated method can then accept a new keyword argument `request_data`,
+    a dictionary of key-value pairs to be passed to the network layer for API
+    calls made by the method.
+
+    :param func:
+        The method to decorate.
```
jmoldow: This needs to be a method of a `Cloneable` subclass. Document that explicitly, and maybe also make that clearer in the param name. Perhaps instead of `func` (which might make one believe that I can decorate anything with it), we can use `method`.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+    :type func:
+        `callable`
+    :return:
+        A wrapped method that can pass extra request data to the network layer.
+    :rtype:
+        :class:`APICallWrapper`
+    """
+    wrapper = APICallWrapper(func)
+    update_wrapper(wrapper, func)
```
jmoldow: I think we could instead do this in `APICallWrapper.__init__`:

```python
def api_call(func):
    return APICallWrapper(func)

class APICallWrapper(object):
    def __init__(self, func_that_makes_an_api_call):
        ...
        update_wrapper(self, func_that_makes_an_api_call)
```

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class APICallWrapper(object):
+
+    def __init__(self, func_that_makes_an_api_call):
+        super(APICallWrapper, self).__init__()
+        self._func_that_makes_an_api_call = func_that_makes_an_api_call
+
+    def __get__(self, _instance, owner):
+        def call(*args, **kwargs):
```
jmoldow: This will cause us to need to construct a new function every time we want to make an API call.

To be more economical, we could try something like this:

```python
class APICallWrapper(object):
    def __init__(self, func_that_makes_an_api_call, instance=None, owner=None):
        ...
        self._instance = instance
        self._owner = owner

    def __get__(self, instance, owner):
        return self.__class__(func_that_makes_an_api_call, instance=instance, owner=owner)

    def __call__(*args, **kwargs):
        ...
```

As another idea, we can consider using https://pypi.python.org/pypi/wrapt to power the decorator, so that we don't have to think at all about the descriptor. Though that would introduce its own slowness.

If we keep this code as-is for now, we should use `@wraps` on the definition of `call`.

boxsdk/session/box_session.py
```diff
@@ -79,12 +79,19 @@ def __init__(self, oauth, network_layer, default_headers=None):
             A dictionary containing default values to be used as headers when this session makes an API request.
         :type default_headers:
             `dict` or None
-        """
+        :param default_network_request_kwargs:
+            A dictionary containing default values to be passed to the network layer
+            when this session makes an API request.
+        :type default_network_request_kwargs:
+            `dict` or None        """
```
jmoldow: """ should be on its own line.

boxsdk/session/box_session.py
```diff
@@ -79,12 +79,19 @@ def __init__(self, oauth, network_layer, default_headers=None):
...
         self._oauth = oauth
```
jmoldow: This method should be calling `super`.

boxsdk/session/box_session.py
```diff
@@ -134,7 +141,10 @@ def with_shared_link(self, shared_link, shared_link_password=None):
         """
         headers = self._default_headers.copy()
         headers.update(get_shared_link_header(shared_link, shared_link_password))
-        return self.__class__(self._oauth, self._network_layer, headers)
+        return self.__class__(self._oauth, self._network_layer, headers, self._default_network_request_kwargs.copy())
+
+    def with_default_network_request_kwargs(self, **kwargs):
```
jmoldow: This should probably be `request_data` (or whatever we call it), not `**kwargs`. One, for consistency and clarity. Two, because using `**kwargs` prevents us from being able to add any keyword arguments to this method (because adding any keyword arguments, which might be used for someone's `request_data`, would break them).

Since this is the only way for objects to specify `request_kwargs`, I think it makes more sense for this to be called `with_network_request_kwargs` (drop the default).

I also feel like the default behavior should be to `update` any existing arguments, not replace them. Otherwise, it's sort of a last-call-wins, and the "default" arguments only have a meaning if there are no more calls to this method.

boxsdk/object/cloneable.py
```diff
@@ -0,0 +1,54 @@
...
+        :param user:
+            The user to impersonate when making API requests.
+        :type user:
+            :class:`User`
+        """
+        return self.clone(self.session.as_user(user))
+
+    def with_shared_link(self, shared_link, shared_link_password):
+        """
+        Returns a new endpoint object with default headers set up to make requests using the shared link for auth.
```
jmoldow: Sphinx needs blank lines between the title, and the rest of the docstring.

This is fine for now, but we should probably do a test run of sphinx, and fix all the errors, before doing a release.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,58 @@
...
+
+class APICallWrapper(object):
+
+    def __init__(self, func_that_makes_an_api_call):
+        super(APICallWrapper, self).__init__()
+        self._func_that_makes_an_api_call = func_that_makes_an_api_call
+        update_wrapper(self, func_that_makes_an_api_call)
+
+    def __get__(self, _instance, owner):
+        @wraps(self._func_that_makes_an_api_call)
```
jmoldow: You added `@wraps` which fixes the name/docstring problem, but just re-copying this here as a note to my future self, in case we need to improve the performance of this code later on:

This will cause us to need to construct a new function every time we want to make an API call.

To be more economical, we could try something like this:

```python
class APICallWrapper(object):
    def __init__(self, func_that_makes_an_api_call, instance=None, owner=None):
        ...
        self._instance = instance
        self._owner = owner

    def __get__(self, instance, owner):
        return self.__class__(func_that_makes_an_api_call, instance=instance, owner=owner)

    def __call__(*args, **kwargs):
        ...
```

As another idea, we can consider using https://pypi.python.org/pypi/wrapt to power the decorator, so that we don't have to think at all about the descriptor. Though that would introduce its own slowness.

boxsdk/util/api_call_decorator.py
```diff
@@ -0,0 +1,58 @@
...
+                    instance = args[0]
+                    args = args[1:]
+                else:
+                    raise TypeError
+            extra_network_parameters = kwargs.pop('extra_network_parameters', None)
+            if extra_network_parameters:
+                # If extra_network_parameters is specified, then clone the instance, and specify the parameters
+                # as the defaults to be used.
+                # pylint: disable=protected-access
+                instance = instance.clone(instance._session.with_default_network_request_kwargs(extra_network_parameters))
```
jmoldow: The point of adding the `session` decorator to the `Cloneable` interface, is that now you can do `instance.session`, and remove the `pylint:disable`.

test/unit/object/test_item.py
```diff
@@ -25,6 +25,21 @@ def test_update_info(test_item_and_response, mock_box_session, etag, if_match_he
     assert update_response.object_id == test_item.object_id
 
 
+def test_update_info_with_default_request_kwargs(test_item_and_response, mock_box_session, mock_box_session_2):
+    # pylint:disable=redefined-outer-name, protected-access
```
jmoldow: I don't think anything in this method requires this?

test/unit/object/test_item.py
```diff
@@ -25,6 +25,21 @@ def test_update_info(test_item_and_response, mock_box_session, etag, if_match_he
...
+    test_item, mock_item_response = test_item_and_response
+    expected_url = test_item.get_url()
+    mock_box_session.with_default_network_request_kwargs.return_value = mock_box_session_2
+    mock_box_session_2.put.return_value = mock_item_response
+    data = {'foo': 'bar', 'baz': {'foo': 'bar'}, 'num': 4}
+    extra_network_parameters = {'timeout': 1}
+    update_response = test_item.update_info(data, extra_network_parameters=extra_network_parameters)
+    mock_box_session.with_default_network_request_kwargs.assert_called_once_with({'timeout': 1})
```
jmoldow: Can just pass `extra_network_parameters` here, instead of re-constructing the `dict`.

jmoldow:
Looks good to me! There are a few minor things I pointed out here, but we can deal with them in another PR. 👍

boxsdk/session/box_session.py
```diff
@@ -134,7 +143,10 @@ def with_shared_link(self, shared_link, shared_link_password=None):
         """
         headers = self._default_headers.copy()
         headers.update(get_shared_link_header(shared_link, shared_link_password))
-        return self.__class__(self._oauth, self._network_layer, headers)
+        return self.__class__(self._oauth, self._network_layer, headers, self._default_network_request_kwargs.copy())
+
+    def with_default_network_request_kwargs(self, extra_network_parameters):
+        return self.__class__(self._oauth, self._network_layer, self._default_headers.copy(), extra_network_parameters)
```
jmoldow: Let's chat about this later:

Since this is the only way for objects to specify `request_kwargs`, I think it makes more sense for this to be called `with_network_request_kwargs` (drop the default).

I also feel like the default behavior should be to `update` any existing arguments, not replace them. Otherwise, it's sort of a last-call-wins, and the "default" arguments only have a meaning if there are no more calls to this method.

boxsdk/util/api_call_decorator.py
jmoldow: Notes for my future self:

This will slow down API calls, since every time you make the call, you'll need to spend time doing this cloning.

That's fine, especially since this feature probably won't be used by the average app. 

We don't necessarily need this for V1 (since it'll only be us using it immediately), but other things I'd be interested in seeing in the future:

- The ability to pass request_data to the decorator. So someone could do

```python
class MyFile(File):
    @api_call(request_data={'foo': 'bar'})
    def get(self, *args, **kwargs):
        ...
```
and then have that clone only happen once per object per method. This would only help if the user happens to be re-using objects for multiple API calls of the same type, which might not be common (so it might never be worth doing this).

- The ability to specify default `request_data` for all methods on an object. So someone could do

```python
class MyFile(File):
    default_request_data = {'foo': 'bar'}
```

or 
```python
class MyFile(File):
    def __init__(self, ..., foo):
        self.default_request_data = {'foo': foo}
```

and have the cloning happen at init time. Then hopefully some method calls might never need `request_data` passed to them at call time.

PR #155, opened 2016-08-01
Register subclasses

kelseymorris95:
Enable custom subclasses of smart objects to be registered correctly with the `Translator`.

### DIFF
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -10,8 +10,8 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 import jwt
 
-from boxsdk.util.compat import total_seconds
 from .oauth2 import OAuth2
+from ..util.compat import total_seconds
 
 
 class JWTAuth(OAuth2):
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -6,13 +6,8 @@ import json
 from ..config import API
 from ..session.box_session import BoxSession
 from ..network.default_network import DefaultNetwork
-from ..object.user import User
-from ..object.folder import Folder
 from ..object.search import Search
 from ..object.events import Events
-from ..object.file import File
-from ..object.group import Group
-from ..object.group_membership import GroupMembership
 from ..util.shared_link import get_shared_link_header
 from ..util.translator import Translator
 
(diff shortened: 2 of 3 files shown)


boxcla:

boxsdk/auth/jwt_auth.py
```diff
@@ -10,8 +10,8 @@
 from cryptography.hazmat.primitives import serialization
 import jwt
 
-from .oauth2 import OAuth2
 from boxsdk.util.compat import total_seconds
```
Jeff-Meadows: may as well change this import to relative as well

Jeff-Meadows:
lgtm 👍

boxsdk/client/client.py
```diff
@@ -143,7 +138,11 @@ def users(self, limit=None, offset=0, filter_term=None):
             params['filter_term'] = filter_term
         box_response = self._session.get(url, params=params)
         response = box_response.json()
-        return [User(self._session, item['id'], item) for item in response['entries']]
+        return [Translator().translate('user')(
```
jmoldow: For better performance, this can be

```python
User = Translator().translate('user')
return [User(...) for ... ]
```

boxsdk/client/client.py
```diff
@@ -240,7 +242,11 @@ def groups(self):
         url = '{0}/groups'.format(API.BASE_API_URL)
         box_response = self._session.get(url)
         response = box_response.json()
-        return [Group(self._session, item['id'], item) for item in response['entries']]
+        return [Translator().translate('group')(
```
jmoldow: For better performance, this can be

```python
Group = Translator().translate('group')
return [Group(...) for ... ]
```

boxsdk/client/client.py
```diff
@@ -342,7 +352,11 @@ def create_user(self, name, login=None, **user_attributes):
             user_attributes['is_platform_access_only'] = True
         box_response = self._session.post(url, data=json.dumps(user_attributes))
         response = box_response.json()
-        return User(self._session, response['id'], response)
+        return Translator().translate(response['type'])(
```
jmoldow: Can be `Translator().translate('user')` for consistency with other endpoints. Though this works too.

(The reason `get_shared_item()` above is special, is because `/shared_items` may return a file or folder.)

boxsdk/client/client.py
```diff
@@ -143,7 +138,12 @@ def users(self, limit=None, offset=0, filter_term=None):
             params['filter_term'] = filter_term
         box_response = self._session.get(url, params=params)
         response = box_response.json()
-        return [User(self._session, item['id'], item) for item in response['entries']]
+        User = Translator().translate('user')
```
jmoldow: Apparently pylint doesn't like uppercase names as variable names -_-

You can either pylint:ignore that, or change it to something like `user_class`.

jmoldow:
👍

PR #156, opened 2016-08-11
Extra exception info

kelseymorris95:
Add `NetworkResponse` to `BoxAPIException`

### DIFF
--- a/boxsdk/exception.py
+++ b/boxsdk/exception.py
@@ -71,6 +71,14 @@ class BoxAPIException(BoxException):
             The context_info returned in the failed response.
         :type context_info:
             `dict`
+        :param request_response:
+            The failed response
+        :type request_response:
+            Requests `Response`
+        :param access_token_used
+            The access token used to make the request
+        :type access_token_used
+            `unicode`
         """
         super(BoxAPIException, self).__init__()
         self._status = status
--- a/boxsdk/exception.py
+++ b/boxsdk/exception.py
@@ -25,7 +25,19 @@ class BoxAPIException(BoxException):
     """
     Exception raised from the box session layer.
     """
-    def __init__(self, status, code=None, message=None, request_id=None, headers=None, url=None, method=None, context_info=None):
+    def __init__(
+        self,
+        status,
+        code=None,
+        message=None,
+        request_id=None,
+        headers=None,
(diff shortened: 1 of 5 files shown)


boxcla:

Jeff-Meadows:
lgtm to push once the other PR is approved.

👍

boxsdk/session/box_session.py
```diff
@@ -248,6 +248,8 @@ def _raise_on_unsuccessful_request(self, network_response, expect_json_response,
                 url=url,
                 method=method,
                 context_info=response_json.get('context_info', None),
+                request_response=network_response.request_response,
```
jmoldow: This is an unnecessary backwards incompatibility, since custom networks might not have `request_response`.

Instead, I would propose that this line be:

`network_response=network_response`

This has the benefits of:
- Is backwards compatible.
- Works with custom subclasses of `NetworkResponse`.
- Still allows the user to access `network_response.request_response`, if `network_response` is an instance of `DefaultNetworkResponse`.
- Not tying us to the `requests` library even further.
- Probably never again needing to add more fields to the exception, since we now include a maximal amount of information.

We could also avoid adding `access_token_used`, since the user could use `exception.network_response.access_token_used` instead. It's a minor difference, we could do either. Having it directly on the object is slightly more clear and direct and less work for the user to access, but is another property that we would have to maintain forever for backwards compatibility.

boxsdk/network/default_network.py
```diff
@@ -71,3 +71,7 @@ def response_as_stream(self):
     def access_token_used(self):
         """Base class override."""
         return self._access_token_used
+
+    @property
+    def request_response(self):
```
jmoldow: Docstring. The one you used above would be fine.

boxsdk/exception.py
```diff
@@ -59,6 +71,14 @@ def __init__(self, status, code=None, message=None, request_id=None, headers=Non
             The context_info returned in the failed response.
         :type context_info:
             `dict`
+        :param request_response:
```
jmoldow: As I said, I think we should use `network_response` instead.

jmoldow:
👍 , waiting for tests to pass.

PR #157, opened 2016-08-15
Add collaboration() method to Client.

kelseymorris95:
For consistency with other objects, can use `client.collaboration('box_id').do_something(...)` the same way we do with `Folder`, `File`, etc.


boxcla:

jmoldow:
👍

PR #158, opened 2016-08-17
Expose _response_object as a property on BaseAPIJSONObject

kelseymorris95:


boxcla:

boxsdk/object/base_api_json_object.py
```diff
@@ -61,3 +61,8 @@ def __repr__(self):
     def _description(self):
         """Return a description of the object if one exists."""
         return ""
+
+    @property
+    def response_object(self):
+        """Return the JSON object representing the object returned from a Box API request."""
```
jmoldow: Document ``:rtype: `dict` ``.

Also, we might want to return a copy of the `dict`, that way if they make modifications to the dict it won't affect the smart object. I will think about which behavior is preferable..

boxsdk/object/base_api_json_object.py
```diff
@@ -61,3 +62,12 @@ def __repr__(self):
...
         return ""
+
+    @property
+    def response_object(self):
+        """
+        Return the JSON object representing the object returned from a Box API request.
+        :rtype:
+            `dict`
+        """
+        return copy.deepcopy(self._response_object)
```
jmoldow: Good idea.

jmoldow:
👍 waiting for tests

PR #159, opened 2016-08-18
Docs update

kelseymorris95:


boxcla:

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
 Upcoming
 ++++++++
 
+New Release?
+++++++++++++++++++
+
+**Breaking Changes**
+
+- ``Events.get_events(...)`` now returns a list of ``Event`` instances rather than a list of ``dicts`` s
```
jmoldow: Just ``dict``, not ``dicts``.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
...
+  ``dict`` .
+
+**Features**
+
+- Added ability to create custom subclasses of SDK objects with ``_item_type`` defined.
+- Added an ``Event`` class.
+
+**Other**
+
+- Added extra information to ``BoxAPIExceptions``s.
```
jmoldow: Remove both s 's.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
...
+- Added ``collaboration()`` method to ``Client``.a
```
jmoldow: Remove trailing a.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
...
 ++++++++
 
+New Release?
+++++++++++++++++++
+
+**Breaking Changes**
+
+- ``Events.get_events(...)`` now returns a list of ``Event`` instances rather than a list of ``dicts`` s
+  representing ``Event`` s.  ``Event`` inherits from ``Mapping`` but will not have all the same capabilities as
+  ``dict`` .
```
jmoldow: Let's add a sub-bullet about how to migrate. Something like:

- .
  - Your code is affected if you use `Events.get_events(...)` and expect a list of `dict` instead of a list of `Mapping`. For example, if you use `__setitem__` (`event['key'] = value`), `update()`, or `copy()`, or if your code depends on the `str` or `repr` of the event. Use of `__getitem__` (`event['key']`), `get`, and other `Mapping` methods is unaffected. See https://docs.python.org/2.7/library/collections.html#collections-abstract-base-classes for methods supported on `Mapping` instances.
  - How to migrate: if you still have the need to treat the events like a `dict`, then you can get a deepcopy of the original `dict` with the new attribute `response_object`.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
...
 ++++++++
 
+New Release?
+++++++++++++++++++
+
+**Breaking Changes**
+
+- ``Events.get_events(...)`` now returns a list of ``Event`` instances rather than a list of ``dicts``
+  representing ``Event`` s.  ``Event`` inherits from ``Mapping`` but will not have all the same capabilities as
+  ``dict`` .
```
jmoldow: Let's add a sub-bullet about how to migrate. Something like:

- .
  - Your code is affected if you use `Events.get_events(...)` and expect a list of `dict` instead of a list of `Mapping`. For example, if you use `__setitem__` (`event['key'] = value`), `update()`, or `copy()`, or if your code depends on the `str` or `repr` of the event. Use of `__getitem__` (`event['key']`), `get`, and other `Mapping` methods is unaffected. See https://docs.python.org/2.7/library/collections.html#collections-abstract-base-classes for methods supported on `Mapping` instances.
  - How to migrate: if you still have the need to treat the events like a `dict`, then you can get a deepcopy of the original `dict` with the new attribute `response_object`.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
 Upcoming
 ++++++++
 
+New Release?
+++++++++++++++++++
+
+**Breaking Changes**
+
+- ``Events.get_events(...)`` now returns a list of ``Event`` instances rather than a list of ``dicts``
```
jmoldow: The s at the end of dict should be outside of the backticks.

HISTORY.rst
```diff
@@ -6,6 +6,30 @@ Release History
...
+
+**Other**
+
+- Added extra information to ``BoxAPIException``.
+- Added ``collaboration()`` method to ``Client``.
+- Reworked the class hierarchy.  Previously, ``BaseEndpoint`` was the parent of ``BaseObject`` which was the parent
+  of all smart objects.  Now ``BaseObject`` is a child of both ``BaseEndpoint`` and ``BaseAPIJSONObject``.
+  ``BaseObject`` is the parent of all objects that are a part of the REST API.  Another subclass of
+  ``BaseAPIJSONObject``, ``APIJSONObject``, was created to represent pseudo-smart objects such as ``Event`` that are not
+  directly accesible through an API endpoint.
```
jmoldow: "accessible"

jmoldow:
👍

PR #166, opened 2016-09-07
Don't log file content when logging Box API responses.

Jeff-Meadows:

PR #172, opened 2016-09-22
Logging refactors and improvements

jmoldow:
Move response logging to a new `LoggingNetworkResponse` class.
This allows us to decide whether to log the response body, based
on whether the caller reads or streams the content. This fixes a
bug introduced by Pull Request #166, where the `LoggingNetwork`
would always omit logging the response body.

Add a `network_response_constructor` property to
`DefaultNetwork`, to make it easier for subclasses to return
their own `NetworkResponse` subclasses. Use this in
`LoggingNetwork`, to return `LoggingNetworkResponse` instances.
Also add it as an optional property on the `Network` interface
class.

In `network_interface.py`, switch some usages of
`@abc.abstractmethod` to `@abc.abstractproperty`. This doesn't
matter on recent versions of Python 3, but there is a difference
between the two on Python 2, plus it makes the interface
clearer. This is a backwards-compatible change, because the SDK
uses these as properties, so any custom implementations would've
needed to implement these with `property` anyway.

Update the logging format strings in `LoggingNetwork`, to make
them more informative and also easier to override (by using
keyword format placeholders instead of positional ones). This is
a breaking change for clients that were overriding these class
attributes.

Add logging for request exceptions in `LoggingNetwork`.

Use the `LoggingNetwork` during functional tests.

Switch some stateful mock fixtures from session scope to
function scope.

### DIFF
--- a/boxsdk/network/logging_network.py
+++ b/boxsdk/network/logging_network.py
@@ -1,8 +1,14 @@
 # coding: utf-8
 
-from __future__ import unicode_literals
+from __future__ import absolute_import, unicode_literals
+
+from functools import partial
 from pprint import pformat
-from boxsdk.network.default_network import DefaultNetwork
+import sys
+
+from six import text_type
+
+from boxsdk.network.default_network import DefaultNetwork, DefaultNetworkResponse
 from boxsdk.util.log import setup_logging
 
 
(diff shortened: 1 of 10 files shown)


boxcla:

boxsdk/network/logging_network.py
```diff
@@ -48,32 +55,216 @@ def _log_request(self, method, url, **kwargs):
...
 
-        :param response: The Box API response.
+        :param method:  The HTTP verb that was used to make the request.
+        :type method:   `unicode`
+        :param url:   The URL for the request.
+        :type url:  `unicode`
+        :param exc_info:  The exception info returned from `sys.exc_info()`.
         """
-        if response.ok:
-            if response.request_response.raw is not None:
```
jmoldow: `response.request_response.raw` is always set (both for `stream=True` and `stream=False`, so this code path was always being followed. cc @Jeff-Meadows

Jeff-Meadows:
👍

PR #178, opened 2016-09-22
Refactors to JWTAuth

### LINKED ISSUE #174: Change authenticate_app_user method
With service accounts, you can use the method to authenticate as a managed user, not just app users. Should the method name be changed to better reflect this?

Also we should changed the name from Box Developer Edition to Box Platform (or even Box service accounts)

jmoldow:
Rename `JWTAuth.authenticate_app_user()` to
`JWTAuth.authenticate_user()`, to reflect the fact that the new
Box service accounts can authenticate as managed users as well
as app users. `JWTAuth.authenticate_app_user()` is retained as
an alias, for backwards-compatability.

Allow a user ID to be passed to `JWTAuth`, instead of a `User`
object. Previously, developers would need to import and manually
construct a `boxsdk.object.user.User`, since `Client.user()`
cannot be used until the `Client` has been constructed (which is
usually done after all authentication is complete).

Allow a user or user ID to be passed to the `JWTAuth`
constructor. This way, `JWTAuth.authenticate_user()` can be
called without any arguments. More importantly, this means that
`JWTAuth.refresh()` can be called immediately after
construction, with no need for a manual call to
`JWTAuth.authenticate_user()`.

When `auto_session_renewal=True`, if there is no access token,
then `BoxSession.request()` will renew the token _before_ making
the request. This saves an API call. And combined with the
above, authentication for `JWTAuth` objects can be done
completely automatically, at the time of first API call.

Document that the `enterprise_id` argument to `JWTAuth` is
allowed to be `None` (and must be `None` if `user` is passed).
Also allow it to be passed to `authenticate_instance()`, so that
it doesn't need to be provided at construction time.

Clean up the README.

Fixes #174.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -323,12 +323,12 @@ These users can then be authenticated:
     ned_auth = JWTAuth(
         client_id='YOUR_CLIENT_ID',
         client_secret='YOUR_CLIENT_SECRET',
-        enterprise_id='YOUR_ENTERPRISE_ID',
+        user=ned_stark_user,
         jwt_key_id='YOUR_JWT_KEY_ID',
         rsa_private_key_file_sys_path='CERT.PEM',
         store_tokens=your_store_tokens_callback_method,
     )
-    ned_auth.authenticate_app_user(ned_stark_user)
+    ned_auth.authenticate_user()
     ned_client = Client(ned_auth)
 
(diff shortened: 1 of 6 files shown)


boxcla:

Jeff-Meadows:
👍

Issue #153, opened 2016-07-26
Can we specify the fields we want when fetch object list?


likel:
Hi, 
When I ingest a list of folder or files, I want to specify the fields to returned, Is that possible?Thanks.

Jeff-Meadows:
Hi, I assume you are talking about the `folder.get_items()` method?

It takes an optional `fields` parameter, which you can use to pass a list of field names.

likel:
@Jeff-Meadows Not only `folder.get_items()`. We can only ingest a list of items via `folder.get_items()` so far. But we have many other  objects such as user, group, comment, task etc. So for these objects, do we have any plan to add the `get_items()` ? I can see we have these APIs from [https://docs.box.com/reference](https://docs.box.com/reference). Thanks.

jmoldow:
Hi @likel 

We will eventually add those in, but we don't have concrete plans yet.

Until somewhat recently, some of those endpoints were incorrectly documented, saying that they returned all results, when actually they were paging endpoints. This caused us to accidentally create non-paging methods for some of those. For example, see https://github.com/box/box-python-sdk/blob/master/boxsdk/client/client.py#L231 , which claims to return all groups. When actually, the endpoint is (now) documented as being a paging endpoint (https://docs.box.com/reference#groups).

Box also has plans to add a better paging mechanism to its endpoints. You can see this in some of the newer paging endpoints, for example https://docs.box.com/reference#get-device-pins . Before this becomes available, we don't want to jump the gun and write a method that would be backwards incompatible with the new paging mechanism.

I think we could probably come up with a good mechanism that supports both limit/offset, and the upcoming paging mechanism. And if we do that, then we can implement these methods whenever we want. But we haven't had a chance to sit down and design that yet.

For now, if you need to use those endpoints, you can construct your own calls, with `client.make_request()`.

likel:
@jarmallen Really thanks your reply. In fact I implemented that functionality via `client.make_request()` now. In other words, I gave up the high level api provided by sdk. I'm happy if you guys can add these API to SDK. I'm also pleased to contribute code. Thanks again for your kindness.

PR #181, opened 2016-11-09
Add task and comment and task assignment api

liketic:
Hi, 

When I ingest data from box api with python sdk, I found there is no comment and task related api, so I added these APIs in this PR. Happy to hear your comments. Thanks.

### DIFF
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -99,6 +99,51 @@ class Client(Cloneable):
         """
         return self.translator.translate('file')(session=self._session, object_id=file_id)
 
+    def comment(self, comment_id):
+        """
+        Initialize a :class:`Comment` object, whose box id is comment_id.
+
+        :param comment_id:
+            The box id of the :class:`Comment` object.
+        :type comment_id:
+            `unicode`
+        :return:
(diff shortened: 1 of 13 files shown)


boxcla:
Verified that @likel has signed the CLA. Thanks for the pull request!

jmoldow:
Thanks for the pull request!

You're going to need to merge master in order to bring your branch up-to-date.

Please add entrypoints for all of these in `boxsdk/client/client.py`. E.g.

```python
def comment(self, comment_id):
    return self.translator.translate('comment')(session=self._session, object_id=comment_id)
```

Also, as I mentioned in https://github.com/box/box-python-sdk/issues/153#issuecomment-236063398 , we haven't yet defined how we want to define a unified API for all paging endpoints. You've defined a few paging endpoints (though, unfortunately, it doesn't look like they are explicitly documented as being paging endpoints in the API reference documentation). I'd prefer if we removed these for now, that way we don't have to make a breaking SDK change later. You should be able to easily define these in your own SDK subclasses, and use them in your application.

likel:
@jmoldow Thanks for your reply. I understand your concern. Thanks.

liketic:
Close this cause no one need it.

PR #183, opened 2016-11-11
Prepare for next 2.0.0 alpha release

jmoldow:
Bump version to 2.0.0a3.

Add new release notes.

gitignore more patterns for Python virtualenv directories.

Start running tests against the development build of Python 3.6
(which is currently in a feature-freeze beta).

`six.with_metaclass()` doesn't work with `enum.EnumMeta` in
Python 3.6, because of logic that was added to
`EnumMeta.__prepare__()`. The six bug actually applies to all
versions of Python 3, but I guess it was never noticed until
now.

This adds our own `with_metaclass()` helper function, with the
fix applied to it. Later I'll submit the same patch to six.
Explanation of the bug and the fix are in the docstring.

jsonpatch is not currently compatible with Python 3.6. If we
can't load it during functional testing, then
`@chaos_utils.patch()` will be a no-op.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ __pycache__/
 # Distribution / packaging
 bin/
 .Python
-env/
+env*/
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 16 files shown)


boxcla:

anthonywee:
👍

Issue #197, opened 2017-03-28
Unable to refresh JWT auth token


GabeLambda:
https://github.com/box/box-python-sdk/blob/master/boxsdk/auth/jwt_auth.py#L195

Callers of this method are expecting a tuple of (`access_token`, `refresh_token`), not just `access_token`. This breaks downstream in https://github.com/box/box-python-sdk/blob/master/boxsdk/session/box_session.py#L196, causing clients which use `JWTAuth` to break on refresh.

Example:
```
File "/usr/local/lib/python3.4/dist-packages/boxsdk/session/box_session.py", line 196, in _renew_session
    new_access_token, _ = self._oauth.refresh(access_token_used) 
ValueError: too many values to unpack (expected 2)
```

Version: 2.0.0a4

maxdebayser:
This fix worked for me

cschloer:
Fixed also worked for me but would rather not using a fork. Any status on this getting into master?

jmoldow:
Fixed via #209

jmoldow:
Also fixed in 1.5 via #217.

I just released a new version, 1.5.4, with this fix.

PR #198, opened 2017-03-28
Fix auth refresh for clients using JWTAuth (#197)

GabeLambda:
This will ensure that the return value is the expected tuple of (access_token, refresh_token), and not just access_token.

### DIFF
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -192,7 +192,7 @@ class JWTAuth(OAuth2):
             data['box_device_id'] = self._box_device_id
         if self._box_device_name:
             data['box_device_name'] = self._box_device_name
-        return self.send_token_request(data, access_token=None, expect_refresh_token=False)[0]
+        return self.send_token_request(data, access_token=None, expect_refresh_token=False)
 
     def authenticate_user(self, user=None):
         """


boxcla:
Hi @GabeLambda, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

GabeLambda:
CLA signed

boxcla:
Verified that @GabeLambda has just signed the CLA. Thanks, and we look forward to your contribution.

jmoldow:
Hi @GabeLambda , thanks for the bug report and the PR.

This particular approach won't work, because this will change the return signature of `authenticate_user()` and `authenticate_instance()` in a way that we don't want.

I was looking into fixing this with this change:

```
diff --git a/boxsdk/auth/jwt_auth.py b/boxsdk/auth/jwt_auth.py
index e536e85..6633c4a 100644
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -291,6 +291,7 @@ def _refresh(self, access_token):
         """
         # pylint:disable=unused-argument
         if self._user_id is None:
-            return self.authenticate_instance()
+            new_access_token = self.authenticate_instance()
         else:
-            return self.authenticate_user()
+            new_access_token = self.authenticate_user()
+        return new_access_token, None
diff --git a/boxsdk/auth/oauth2.py b/boxsdk/auth/oauth2.py
index 6b0e9d5..ccfd494 100644
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -206,11 +206,10 @@ def refresh(self, access_token_to_refresh):
                 # If the active access token is the same as the token needs to
                 # be refreshed, or if we don't currently have any active access
                 # token, we make the request to refresh the token.
-                return self._refresh(access_token_to_refresh)
-            else:
-                # If the active access token (self._access_token) is not the same as the token needs to be refreshed,
-                # it means the expired token has already been refreshed. Simply return the current active tokens.
-                return access_token, refresh_token
+                access_token, refresh_token = self._refresh(access_token_to_refresh)
+            # Else, if the active access token (self._access_token) is not the same as the token needs to be refreshed,
+            # it means the expired token has already been refreshed. Simply return the current active tokens.
+        return access_token, refresh_token
```

I hadn't committed that yet because I was also working on getting the `OAuth2` tests to run against `JWTAuth` as well (this bug would've been caught if we had been doing that).

GabeLambda:
Fair enough. I had assumed that those signatures should look the same as `_refresh`'s, given that they were being returned directly.

jmoldow:
Since they are public methods, we should assume that users may be calling them directly and using the return values somehow.

We _could_ change the signature, and it'd be a breaking change that we could put into the upcoming `2.0.0` release. But it seems like an unnecessary breaking change. Since the second element of the tuple will always be `None`, even in subclasses, the method would be less useful if it returned a tuple.

Jeff-Meadows:
Opened @jmoldow's suggestion as https://github.com/box/box-python-sdk/pull/209

Issue #113, opened 2016-02-10
Collaboration API


Devouass:
Hello all.
First thanks for the job done here. Very helpfull.
I have a question, is it plan to adress collaboration's actions?
I notice than with the client.make_request function, it is possible to make any requests, but an dedicated API is very usefull.
This is not an issue, more a question about what are the futur things planned. (Maybe there is a dedicated file or place somewhere which references next improvements ?).
Thanks

Jeff-Meadows:
We don't exactly have a roadmap. The Box API continually adds new features, and the SDK adds new features over time as well. We accept pull requests for people wishing to add features, or we add them when we have time. What part of the collab API are you hoping to see added to the SDK?

Devouass:
Currently I'm looking on how associate groups and folder or files, for rights management.
And as I understand the Box API, this management is allowed using collaboration API.
With the SDK, we can create groups, add users in groups, create/upload/delete files and folder. But I did not see anything about associate groups and folder. (And, maybe i'm wrong, but I did not see anything about create new user).
Cleverly, As I mentionned in the post before, you openned an access to the network layer with the client.make_request method.
But I don't know if what I want to do is very specific and not very usefull for most people, or if those features should be added in the SDK.

Jeff-Meadows:
You can create a user with `client.create_user()`: http://box-python-sdk.readthedocs.org/en/latest/boxsdk.html#boxsdk.client.Client.create_user

And yes, you associate groups to folders via collaborations (currently, collabs only work at the folder level, so there is no such thing as a file collaboration). You do this via `folder.add_collaborator()`: http://box-python-sdk.readthedocs.org/en/latest/boxsdk.object.html#boxsdk.object.folder.Folder.add_collaborator

The collaborator may be a `User`, a `Group`, or just an email address.

Devouass:
Thanks for your answers Jeff.
I apologize, I think everything I need is already implemented, I did not see it before...
Thanks for the answers, and for the work done.

Devouass:
Hello Jeff.
I search for a feature, but I'm not sure this feature is implemented on the SDK.
I deal with groups now, and in the API, I can have a generator for group's membership.
I see in the API that we can also retrieve the collaboration concerned by this group, using the request "BOX/groups/[group_id]/collaborations"
It is the same with a folder approach ("BOX/folders/[folder_id]/collaborations")
[...]

PR #206, opened 2017-05-25
Add get_collaborators() to Folder object. 

nddipiazza:
Addresses #113. Adding function get_collaborators in the Folder class to bring sdk closer to Box API.


mattwiller:
We added [`item.get_collaborations()`](https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.item.Item.get_collaborations) to the SDK, which supersedes this change

PR #208, opened 2017-06-21
Add Collection object and collections()/get_favorites() to Client

josephroque:
Use the Client object to get the user's collections (which just returns a list containing the singular Favorites collection). I also added a convenience method for getting favorite items directly.

Collections (i.e. just the Favorites collection) can be used to retrieve items, as well as add or remove an item from the collection. There is also `Item.add_to_collection` and `Item.remove_from_collection` which do just as they say. Note that `Item.remove_from_collection` and `Collection.remove_item` remove an item from **all collections**, since there's no way to get all of the collections an item belongs to and just remove one, as well as there's only the Favorites collection.


boxcla:

PR #209, opened 2017-06-21
Fix JWT Token Refresh

Jeff-Meadows:


boxcla:

PSutton3756:
Any word on when this will be merged into master? Seems like this is a pretty common issue and it would suck to have to push a separate branch into a production environment if it can be avoided.

Issue #140, opened 2016-06-16
Auto-translate all Box REST API objects


jmoldow:
Box REST API endpoints usually return JSON. Most of these are either objects (with a `type` field and an `id` field), "pseudo-objects" (something with a `type`, but not an `id`, e.g. an `event`), or collections (e.g. the responses from `/folder/:id:/items` or `/events`). Furthermore, they will all usually have references to other objects (e.g. the `parent` of an item, the `source` of an `event`, or the `entries` of an `/events` response).

By default, all of these will be returned from the SDK as JSON `dict` objects. The only places were we do translations into the smart instances of `BaseObject` subclasses is in locations where we explicitly do the translation ourselves:
- https://github.com/box/box-python-sdk/blob/3ba6cf1fe50f2f9b3c273d96375dfdc98ea87d3c/boxsdk/object/group_membership.py#L73
- https://github.com/box/box-python-sdk/blob/3ba6cf1fe50f2f9b3c273d96375dfdc98ea87d3c/boxsdk/object/search.py#L232
- https://github.com/box/box-python-sdk/blob/3ba6cf1fe50f2f9b3c273d96375dfdc98ea87d3c/boxsdk/object/base_object.py#L240
- https://github.com/box/box-python-sdk/blob/3ba6cf1fe50f2f9b3c273d96375dfdc98ea87d3c/boxsdk/client/client.py#L292
- https://github.com/box/box-python-sdk/blob/3ba6cf1fe50f2f9b3c273d96375dfdc98ea87d3c/boxsdk/object/folder.py#L155

It'd be great if this translation could happen automatically, for arbitrarily nested objects, and remove most of these manual translations.

Since pseudo-objects and collections are not REST objects, they cannot be instances of `BaseObject`. So this functionality cannot be written directly into `BaseObject`. Instead, we can introduce a new mixin with this functionality (perhaps called `AutomaticTranslatorMixin`). We could implement it in one of two ways:
- During `__init__`, traverse the entire JSON structure and translate everything.
- Override `__getattr__` and `__getitem__` to do translation on demand.

A new class, `BaseAPIJSONObject`, would inherit from `AutomaticTranslatorMixin`. And `BaseObject` would inherit from `BaseAPIJSONObject`.

Pseudo-objects and collections would need to be translated into `BaseAPIJSONObject` instances, not `BaseObject` instances. However, 
[...]

PR #210, opened 2017-07-05
Add new Recents API to client

josephroque:
See: https://developer.box.com/v2.0/reference#recent-item-object

Use the Client to retrieve a user's recently accessed items on Box.

`RecentItem` class allows the translator to correctly instantiate the item returned by the API, which has the field `type='recent_item'`. This provides some additional fields, and access to the underlying File/Folder as the property `RecentItem.item`

### DIFF
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -10,6 +10,7 @@ from ..object.cloneable import Cloneable
 from ..util.api_call_decorator import api_call
 from ..object.search import Search
 from ..object.events import Events
+from ..pagination.marker_based_object_collection import MarkerBasedObjectCollection
 from ..util.shared_link import get_shared_link_header
 
 
@@ -317,37 +318,38 @@ class Client(Cloneable):
         )
 
     @api_call
-    def get_recent_items(self, limit=100, offset=0, fields=None):
+    def get_recent_items(self, limit=None, marker=None, fields=None):
         """
         Get the user's recently accessed items.
 
         :param: limit
-            The maximum number of items to return
+            The maximum number of items to return. If limit is set to None, then the default
+            limit (returned by Box in the response) is used. See https://developer.box.com/reference#get-recent-items
+            for default.
         :type: limit
-            `int`
-        :param offset:
+            `int` or None
+        :param marker:
             The index at which to start returning items.
-        :type offset:
-            `int`
+        :type marker:
+            `str` or None
         :param fields:
-            List of fields to request.
+            List of fields to request on the file or folder which the `RecentItem` references.
         :type fields:
             `Iterable` of `unicode`
         :returns:
-            A list of the user's recently accessed items.
+            An iterator on the user's recent items
         :rtype:
-            `list` of :class:`Item`
+            :class:`MarkerBasedObjectCollection`
         """
-        url = '{0}/recent_items'.format(API.BASE_API_URL)
-        params = {
-            'limit': limit,
-            'offset': offset,
-        }
-        if fields:
-            params['fields'] = ','.join(fields)
-        box_response = self._session.get(url, params=params)
-        response = box_response.json()
-        return [self.translator.translate(item['type'])(session=self._session, response_object=item) for item in response['entries']]
+        return MarkerBasedObjectCollection(
+            self.session,
+            self.get_url('recent_items'),
+            limit=limit,
+            fields=fields,
+            marker=marker,
+            return_full_pages=True,
+            supports_limit_offset_paging=False,
+        )
 
     @api_call
     def get_shared_item(self, shared_link, password=None):
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -318,7 +318,7 @@ class Client(Cloneable):
         )
 
     @api_call
-    def get_recent_items(self, limit=None, marker=None, fields=None):
+    def get_recent_items(self, limit=None, marker=None, fields=None, **kwargs):
         """
         Get the user's recently accessed items.
 
@@ -347,8 +347,7 @@ class Client(Cloneable):
             limit=limit,
             fields=fields,
             marker=marker,
-            return_full_pages=True,
-            supports_limit_offset_paging=False,
+            **kwargs
         )
 
     @api_call
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -347,7 +347,7 @@ class Client(Cloneable):
             params['fields'] = ','.join(fields)
         box_response = self._session.get(url, params=params)
         response = box_response.json()
-        return [self.translator.translate(item['type'])(self._session, item) for item in response['entries']]
+        return [self.translator.translate(item['type'])(session=self._session, response_object=item) for item in response['entries']]
 
     @api_call
     def get_shared_item(self, shared_link, password=None):
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -317,6 +317,39 @@ class Client(Cloneable):
         )
 
     @api_call
+    def get_recent_items(self, limit=100, offset=0, fields=None):
+        """
+        Get the user's recently accessed items.
+
+        :param: limit
+            The maximum number of items to return
+        :type: limit
+            `int`
+        :param offset:
+            The index at which to start returning items.
+        :type offset:
+            `int`
+        :param fields:
+            List of fields to request.
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            A list of the user's recently accessed items.
+        :rtype:
+            `list` of :class:`Item`
+        """
+        url = '{0}/recent_items'.format(API.BASE_API_URL)
+        params = {
+            'limit': limit,
+            'offset': offset,
+        }
+        if fields:
+            params['fields'] = ','.join(fields)
+        box_response = self._session.get(url, params=params)
+        response = box_response.json()
+        return [self.translator.translate(item['type'])(self._session, item) for item in response['entries']]
+
+    @api_call
     def get_shared_item(self, shared_link, password=None):
         """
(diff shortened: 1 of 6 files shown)


boxcla:

boxsdk/client/client.py
```diff
@@ -317,6 +317,39 @@ def create_group(self, name):
         )
 
     @api_call
+    def get_recent_items(self, limit=100, offset=0, fields=None):
```
Jeff-Meadows: It appears this endpoint uses marker based paging, not limit/offset.

boxsdk/client/client.py
```diff
@@ -317,6 +317,39 @@ def create_group(self, name):
...
+        :type offset:
+            `int`
+        :param fields:
+            List of fields to request.
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            A list of the user's recently accessed items.
+        :rtype:
+            `list` of :class:`Item`
```
Jeff-Meadows: this returns a list of `RecentItem`, right?

boxsdk/client/client.py
```diff
@@ -317,6 +317,39 @@ def create_group(self, name):
...
+        :param: limit
+            The maximum number of items to return
+        :type: limit
+            `int`
+        :param offset:
+            The index at which to start returning items.
+        :type offset:
+            `int`
+        :param fields:
+            List of fields to request.
```
Jeff-Meadows: Mention that the fields apply to the recent item object, not to the underlying file or folder.

boxsdk/client/client.py
```diff
@@ -317,6 +317,39 @@ def create_group(self, name):
...
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            A list of the user's recently accessed items.
+        :rtype:
+            `list` of :class:`Item`
+        """
+        url = '{0}/recent_items'.format(API.BASE_API_URL)
```
Jeff-Meadows: this can be `self.get_url('recent_items')`

josephroque: Everything else in the file follows the format used here. Should I still change it?

boxsdk/client/client.py
josephroque: Actually, these do apply to the underlying file or folder. I'll mention that. The recent_item type always has the same fields, but returns the 'mini' file/folder format by default, so this is to request more fields. It's similarly unclear in the API documentation, but that is what it does in practice. I'll update this to be more clear.

boxsdk/client/client.py
josephroque: 👍 waiting on #212 to rebase

boxsdk/client/client.py
Jeff-Meadows: Oh, cool.

boxsdk/client/client.py
Jeff-Meadows: Up to you, I guess. Most of the other things in this file predate being able to use `self.get_url`

boxsdk/pagination/page.py
```diff
@@ -61,7 +61,7 @@ def __getitem__(self, key):
         item_class = self._translator.translate(item_json['type'])
         item = item_class(
             session=self._session,
-            object_id=item_json['id'],
+            object_id=item_json['id'] if 'id' in item_json else None,
```
josephroque: Note that `RecentItem` is not a subclass of `BaseObject` because most of the functionality provided by `BaseObject` does not apply to to recent items (o way to GET individual recent items, no way to update). Finally, recent items from the API do not have unique IDs associated with them, so there's no way to get `item_json['id']`

boxsdk/client/client.py
```diff
@@ -317,6 +318,40 @@ def create_group(self, name):
         )
 
     @api_call
+    def get_recent_items(self, limit=None, marker=None, fields=None):
```
jmoldow: We should add `**kwargs`, and pass that to `MarkerBasedObjectCollection`. And then remove `return_full_pages=True` below, since that's a decision that the client, not the SDK, should make.

Depending on your personal preference, you could remove `limit`, `marker`, and `fields`, and leave them all for `**kwargs`. Or you can leave it as-is.

boxsdk/client/client.py
```diff
@@ -317,6 +318,40 @@ def create_group(self, name):
...
+            :class:`MarkerBasedObjectCollection`
+        """
+        return MarkerBasedObjectCollection(
+            self.session,
+            self.get_url('recent_items'),
+            limit=limit,
+            fields=fields,
+            marker=marker,
+            return_full_pages=True,
+            supports_limit_offset_paging=False,
```
jmoldow: I don't know if this is necessary, I believe it defaults to `False`.

boxsdk/object/recent_item.py
```diff
@@ -0,0 +1,45 @@
...
+            :class:`Item`
+        """
+        item = self._response_object['item']
+        return self.translator.translate(item['type'])(
+            session=self._session,
+            object_id=item['id'],
+            response_object=item,
+        )
+
+    def get_url(self, *args):
```
jmoldow: Is this needed? As you said, this isn't a `BaseObject`, so we'll never be calling API endpoints for this object.

boxsdk/object/recent_item.py
```diff
@@ -0,0 +1,45 @@
...
+from .base_endpoint import BaseEndpoint
+from .base_api_json_object import BaseAPIJSONObject
+
+
+class RecentItem(BaseEndpoint, BaseAPIJSONObject):
+    """Represents a single recent item accessed by a Box user."""
+
+    _item_type = 'recent_item'
+
+    def __init__(self, session, response_object=None, **kwargs):
```
jmoldow: We shouldn't accept `**kwargs` unless we're willing to do something with it.

Doing this is just a hack around the fact that we're passing `object_id` to it..

We should solve that problem a different way.

And if we remove `**kwargs`, then you can actually just remove this method override altogether, since it isn't doing anything other than calling `super`.

boxsdk/pagination/page.py
jmoldow: Maybe we can do something like this:

```python
item_class = self._translator.translate(item_json['type'])
kwargs = {}
if issubclass(item_class, BaseEndpoint):
    kwargs['session'] = self._session
if issubclass(item_class, BaseObject):
    kwargs['object_id'] = item_json['id'] if 'id' in item_json else None
item = item_class(response_object=item_json, **kwargs)
```

It's not great, but it should work, for now. Eventually we'll make something nicer than this for #140 .

test/unit/client/test_client.py
```diff
@@ -61,6 +62,11 @@ def folder_id():
 
 
 @pytest.fixture(scope='module')
+def marker_id():
+    return 'marker_1'
+
+
+@pytest.fixture(scope='module')
```
jmoldow: Won't this break if it is a module fixture? Because the `Mock` will retain its call spec between tests if we do this.

test/unit/client/test_client.py
```diff
@@ -145,6 +151,19 @@ def search_response(file_id, folder_id):
     return mock_network_response
 
 
+@pytest.fixture(scope='module')
```
jmoldow: Again, this doesn't seem like something that should be module scoped.

test/unit/client/test_client.py
```diff
@@ -145,6 +151,19 @@ def search_response(file_id, folder_id):
...
+def recent_items_response(file_id, marker_id):
+    mock_network_response = Mock(DefaultNetworkResponse)
+    mock_network_response.json.return_value = {
+        'entries': [
+            {'type': 'recent_item', 'item': {'type': 'file', 'id': file_id}}
+        ],
+        'next_marker': marker_id,
```
jmoldow: Providing a `next_marker` implies that there are more results coming. So if you try paging, it'll try to make a second API call, which doesn't seem like what you want.

test/unit/client/test_client.py
josephroque: This `user_response` object was pre-existing, git just reports me as adding it. I added the lines above, with the @pytext.fixture from there. So for the comment on scope below, I followed suit from this line you're commenting on here. Additionally, I believe most of these fixtures are used in a read-only sense, so reuse of them in the tests is inconsequential.

test/unit/client/test_client.py
josephroque: See above

boxsdk/pagination/page.py
josephroque: I think this is what's going to be necessary - solves the `**kwargs` in RecentItem as well.

test/unit/object/test_recent_item.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_init_recent_item(mock_box_session, mock_object_id):
+    recent_item = RecentItem(
+        session=mock_box_session,
+        response_object={
+            "type": "recent_item",
+            "item": {"type": "file", "id": mock_object_id}
+        })
+    assert recent_item['type'] == 'recent_item'
+    assert recent_item.item.object_id == mock_object_id
```
jmoldow: Can you also `assert recent_item.item.session is mock_box_session`?

boxsdk/client/client.py
```diff
@@ -317,6 +318,39 @@ def create_group(self, name):
         )
 
     @api_call
+    def get_recent_items(self, limit=None, marker=None, fields=None, **kwargs):
```
jmoldow: In docstring, document that `**kwargs` are passed to `MarkerBasedObjectCollection`.

We could also call this `**collection_kwargs` or `**paging_kwargs`.

josephroque: Cool, I was going to do this, but noticed it wasn't done that way somewhere else, so I thought it might be like, a standard.

PR #211, opened 2017-07-05
July 2017, Pylint/Test environment fixes

josephroque:
Since the most recent successful merge, a number of minor version changes have been made to various dependencies, which have caused a number of linting errors to appear, and for certain configurations to no longer be supported.

In general, this PR fixes pylint errors that make the most sense to fix (no-else-return and len-as-condition) and ignores those that would be API breaking (arguments-differ, esp. in `boxsdk/object` files)

Pylint also uses isort to order dependencies now, which causes some test files to fail linting. These were fixed simply by invoking `isort <filename>`

Sphinx, used for doc generation, no longer supports Python 2.6. A decision needed to be made to either drop Python2.6 testing, or lock Sphinx to version <1.6 (when the typings dependency was added, which requires Python2.7+). I felt uncomfortable removing Python2.6 testing, and the documentation was last run with Sphinx 1.5.1 I believe, so that seemed like the safest decision (locking Sphinx).

Finally, pypy has been updated to 5.7.0 for the tox testing env. This was because the cryptography dependency has undergone a number of updates, from 0.9.2 (the last tested version) up to 1.9. As of 1.0, cryptography no longer supports pypy < 5.3. Either pypy must be upgraded or cryptography locked to <1.0. I figured updates the cryptography would outweigh the decision to keep pypy 4.0.0

### DIFF
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -36,12 +36,12 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
             return '{0} ({1})'.format(self._object_id, self.name)  # pylint:disable=no-member
         return '{0}'.format(self._object_id)
 
-    # pylint:disable=arguments-differ
     def get_url(self, *args):
         """
         Base class override.
         Return the given object's URL, appending any optional parts as specified by args.
         """
+        # pylint:disable=arguments-differ
         return super(BaseObject, self).get_url('{0}s'.format(self._item_type), self._object_id, *args)
 
     def get_type_url(self):
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -34,9 +34,9 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
         """Base class override.  Return a description for the object."""
         if 'name' in self._response_object:
             return '{0} ({1})'.format(self._object_id, self.name)  # pylint:disable=no-member
-        else:
-            return '{0}'.format(self._object_id)
+        return '{0}'.format(self._object_id)
 
+    # pylint:disable=arguments-differ
     def get_url(self, *args):
         """
         Base class override.
--- a/boxsdk/object/events.py
+++ b/boxsdk/object/events.py
@@ -53,6 +53,7 @@ class EnterpriseEventsStreamType(EventsStreamType):
 class Events(BaseEndpoint):
     """Box API endpoint for subscribing to changes in a Box account."""
 
+    # pylint:disable=arguments-differ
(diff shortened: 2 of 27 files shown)


boxcla:

josephroque:
Changed the Pypy update from 5.7.0 to 2.7-5.8.0. Apparently 5.7.0 broke support with Ubuntu 12.04, which Travis tends to use to build projects. Chose 2.7-5.8.0 over 3.5-5.8.0 since 2.7 seems to be the de facto choice for Python version for testing for the rest of the commands.

boxsdk/object/base_object.py
```diff
@@ -34,9 +34,9 @@ def _description(self):
         """Base class override.  Return a description for the object."""
         if 'name' in self._response_object:
             return '{0} ({1})'.format(self._object_id, self.name)  # pylint:disable=no-member
-        else:
-            return '{0}'.format(self._object_id)
+        return '{0}'.format(self._object_id)
 
+    # pylint:disable=arguments-differ
```
Jeff-Meadows: if it's just for this method, please put the pylint directive inside the method, like so:

    def get_url(self, *args):
        """Docstring"""
        # pylint:disable---

boxsdk/object/events.py
```diff
@@ -53,6 +53,7 @@ class EnterpriseEventsStreamType(EventsStreamType):
 class Events(BaseEndpoint):
     """Box API endpoint for subscribing to changes in a Box account."""
 
+    # pylint:disable=arguments-differ
```
Jeff-Meadows: same as above

boxsdk/object/folder.py
```diff
@@ -369,6 +369,7 @@ def add_collaborator(self, collaborator, role, notify=False):
             response_object=collaboration_response,
         )
 
+    # pylint:disable=arguments-differ
```
Jeff-Meadows: same as above

test/functional/conftest.py
```diff
@@ -3,20 +3,20 @@
 from __future__ import unicode_literals
 
 import re
+from test.functional.mock_box.box import Box
```
Jeff-Meadows: not sure what the linter is doing here, but it's causing you to make a bunch of changes that aren't what we want.

we want our imports to be grouped into 1) builtins, 2) 3rd party, 3) things from boxsdk (or test)

you may need to change the pylintrc file so it agrees with our style

josephroque: Pylint/isort isn't aware by default that 'test...' would be a "first party" import. All it took was a declaration in setup.cfg

PR #212, opened 2017-07-12
Limit/Offset and Marker Based pagination

josephroque:
Enable marker based pagination in the SDK, explicitly required for the [new Recents API](https://developer.box.com/v2.0/reference#get-recent-items)

Recents isn't compatible with the typical [limit/offset paging](https://developer.box.com/v2.0/reference#offset-based-paging) that most other endpoints use, and so [marker-based paging](https://developer.box.com/v2.0/reference#marker-based-paging) must be enabled. This PR adds explicit classes to support Limit/Offset pagination, as well as Marker pagination.


boxcla:

PR #218, opened 2017-07-27
APICallWrapper.__get__ must return a bound method

jmoldow:
This magic method was previously returning a plain, unbound
function. This tricks some introspection code, which expects it
to be a bound method.

Also add unit test coverage for that module.

Bump version to 2.0.0a6.

### DIFF
--- a/test/unit/util/test_api_call_decorator.py
+++ b/test/unit/util/test_api_call_decorator.py
@@ -27,6 +27,7 @@ def api_call_method_fixture(api_call_result):
 def cloneable_subclass_with_api_call_method(api_call_method):
     api_call_method_fixture = api_call_method
 
+    # pylint:disable=abstract-method
     class CloneableSubclass(Cloneable):
         api_call_method = api_call_method_fixture
 
@@ -36,6 +37,7 @@ def cloneable_subclass_with_api_call_method(api_call_method):
 @pytest.fixture
 def mock_cloneable(cloneable_subclass_with_api_call_method):
 
+    # pylint:disable=abstract-method
     class MockCloneable(cloneable_subclass_with_api_call_method, NonCallableMock):
         pass
 
@@ -75,6 +77,7 @@ def test_api_call_decorated_method_must_be_a_cloneable_method():
 
 
 def test_api_call_decorated_method_must_be_bound_to_an_instance_of_the_owner(mock_cloneable, api_call_method):
+    # pylint:disable=abstract-method
     class CloneableSubclass2(Cloneable):
         pass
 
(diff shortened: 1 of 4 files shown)


boxcla:

test/unit/util/test_api_call_decorator.py
```diff
@@ -0,0 +1,116 @@
...
+    assert api_call_method.__get__(mock_cloneable, Cloneable) is not api_call_method
+    assert api_call_method.__get__(mock_cloneable, Cloneable).__self__ is mock_cloneable
+    assert mock_cloneable.api_call_method is not api_call_method
+    assert mock_cloneable.api_call_method.__self__ is mock_cloneable
+
+
+def test_api_call_decorated_method_delegates_to_wrapped_method(mock_cloneable, api_call_result):
+    args = (1, 2, 'ƒøø', 'bar')
+    kwargs = {'bar': 'ƒøø'}
+    assert mock_cloneable.api_call_method(*args, **kwargs) == (mock_cloneable, args, kwargs, api_call_result)
```
Jeff-Meadows: you could take the `api_call_method` fixture, and call it instead of encoding here the functionality of the method

    assert mock_cloneable.api_call_method(*args, **kwargs) == api_call_method(mock_cloneable, *args, **kwargs)

test/unit/util/test_api_call_decorator.py
```diff
@@ -0,0 +1,116 @@
...
+
+def test_api_call_decorated_method_can_be_called_as_an_unbound_method_with_an_instance_as_the_first_argument(
+        mock_cloneable,
+        api_call_result,
+        cloneable_subclass_with_api_call_method,
+):
+    args = (1, 2, 'ƒøø', 'bar')
+    kwargs = {'bar': 'ƒøø'}
+    api_call_method = cloneable_subclass_with_api_call_method.api_call_method
+    assert api_call_method(mock_cloneable, *args, **kwargs) == (mock_cloneable, args, kwargs, api_call_result)
```
Jeff-Meadows: ditto here

test/unit/util/test_api_call_decorator.py
```diff
@@ -0,0 +1,116 @@
...
+    def func():
+        pass
+
+    with pytest.raises(TypeError):
+        func()
+
+
+def test_api_call_decorated_method_must_be_a_cloneable_method():
+
+    class Cls(object):
```
jmoldow: Done.

test/unit/util/test_api_call_decorator.py
jmoldow: I would prefer to leave this as-is.

If I do what you suggest, then we'll be passing through `@api_call` on both the LHS and RHS of the equality.

I could expose a fixture for the unbound method, but I'd rather not do that. I think it's fine as-is.

PR #219, opened 2017-08-03
Correctly handle upload responses

### LINKED ISSUE #150: Mock Box does not response correctly to File.update_contents()
The structure of an API response during a call to `File.update_contents()` is documented here:
https://docs.box.com/reference#upload-a-new-version-of-a-file

The documented structure for a response is:
`{
    "total_count": 1,
    "entries": [
        {
            "type": "file",
            ...,
        }
    ]
}`

The current response received from Mock Box for the same API request in test/functional/test_file_upload_update_download.py is:
`{
"type": "file",
...,
}`

`Folder` also has an `upload()` method.  This method should receive a response object with the same structure as `File.update_content()`'s documented version, as documented below:
https://docs.box.com/reference#upload-a-file

In this case, a `Folder` object is built and returned with `box_response['entries'][0]` as its `response_object` member.

In `File.update_contents()`, the `response_object` member is set to the entire response instead, leading to an inconsistency in the way these two similar operations are handled.

jmoldow:
Upload responses look like {"entries": [{"type": "file", ...}]},
not {"type": "file", ...}. However, the code for uploading a new
file version was incorrectly assuming the latter, causing it to
return File({"entries": [{"type": "file", ...}]}) instead of
File({"type": "file", ...}).

Now we will expand the "entries", and grab only the file object.

We'll also be future-proof, in case a future version of the API
starts returning {"type": "file", ...}.

Fixes #150.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -31,6 +31,12 @@ Release History
   automatically backwards-compatibile (as long as there aren't any changes to
   change/remove any of the keywords).
 
+- ``File.update_contents()`` and ``File.update_contents_with_stream()`` now
+  correctly return a ``File`` object with the correct internal JSON structure.
+  Previously it would return a ``File`` object where the file JSON is hidden
+  inside ``file['entries'][0]``. This is a bugfix, but will be a breaking
+  change for any clients that have already written code to handle the bug.
+
 **Features**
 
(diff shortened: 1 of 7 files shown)


boxcla:

PR #222, opened 2017-08-12
Support for passing RSA private key to JWTAuth

### LINKED ISSUE #177: Allow JWTAuth to be constructed without a private key sys path
It should be possible to pass an arbitrary `io` object, or possibly a byte string.

That way, users have a lot more flexibility with how they provide the private key to the `JWTAuth` object.

jmoldow:
This allows a `JWTAuth` object to be initialized, without
requiring that the key be available to the process on disk.

Fixes #177.


boxcla:

Issue #175, opened 2016-09-22
Add ability for OAuth2 class to auto-revoke tokens


jmoldow:
In certain use cases (e.g. not using `RemoteManagedOAuth2`, not using `store_tokens`), the client might not have a need for the access tokens / refresh tokens. For maximum security, it would be best if such clients called `OAuth2.revoke()` at the end of their program, so that the tokens becomes useless, in case an adversary was able to get them.

But the client may not remember to call `revoke()` in all relevant locations, so it could be beneficial to allow the object to auto-revoke its tokens.

I propose two mechanisms for this:
- Add `__enter__` and `__exit__` methods to `OAuth2`, making into a context manager that can be used in `with` statements, which would auto-revoke the token at exit.
- Add an optional flag to the `OAuth2` constructor. When enabled, the `__del__` destructor would call `revoke()`.

These would both be optional (the object would continue to work without being used as a context manager, and the constructor flag would default to `False`). One, for backwards-compatability. Two, because neither of these mechanisms is guaranteed to work (e.g. the process might be forcibly terminated; the computer may be offline at the time of the `revoke()` call), so we'll leave it up to the developer to decide how best to handle revoking. Three, because there are probably situations where revoking at exit is not desired, so we don't want to do that by default.

jmoldow:
A version of this is implemented in #224 . However, it differs from the original proposal:

- Rather than implementing `__enter__` and `__exit__`, I implemented a context manager method called `closing()` (similar to `contextlib.closing()`). The reason for doing it this way was two-fold:
  - Because the logic I wanted to write was easier to write using `@contextmanager`, as opposed to writing inside `__exit__`.
  - Since it is a method, it allows the caller to pass parameters that customize the behavior of the context manager.

- I also opted not to implement `__del__`. I tried it out, but it seemed to produce problems. Trying to make a network call at destruction time can, it seems, hang a program, which isn't desirable behavior. The context manager should instead be used by anyone that wants auto-revoking behavior.

PR #224, opened 2017-08-22
Add functionality for auto-revoking auth tokens

jmoldow:
Auth objects can now be closed, which prevents them from being
used to request new tokens. This will also revoke any existing
tokens. Also introduces a `closing()` context manager method,
which will auto-close the auth object on exit. Clients can use
this to make sure that their tokens don't live longer than they
need them for.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -2,10 +2,13 @@
 
 from __future__ import unicode_literals
 
+from contextlib import contextmanager
 from threading import Lock
 import random
 import string  # pylint:disable=deprecated-module
+import sys
 
+import six
 from six.moves.urllib.parse import urlencode, urlunsplit  # pylint:disable=import-error,no-name-in-module
 
 from boxsdk.network.default_network import DefaultNetwork
@@ -16,6 +19,11 @@ from boxsdk.exception import BoxOAuthException
 class OAuth2(object):
     """
     Responsible for handling OAuth2 for the Box API. Can authenticate and refresh tokens.
+
+    Can be used as a closeable resource, similar to a file. When `close()` is
+    called, the current tokens are revoked, and the object is put into a state
+    where it can no longer request new tokens. This action can also be managed
+    with the `closing()` context manager method.
     """
 
     def __init__(
@@ -77,6 +85,7 @@ class OAuth2(object):
         self._refresh_lock = refresh_lock or Lock()
         self._box_device_id = box_device_id
         self._box_device_name = box_device_name
+        self._closed = False
 
     @property
     def access_token(self):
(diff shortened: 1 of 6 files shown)


boxcla:

boxsdk/auth/oauth2.py
```diff
@@ -320,3 +342,72 @@ def revoke(self):
...
+            self.close(**close_kwargs)
+        except Exception:
+            exc_infos.append(sys.exc_info())
+
+        if exc_infos:
+            six.reraise(*exc_infos[0])
+
+    def _check_closed(self):
+        if self.closed:
+            raise ValueError("operation on a closed auth object")
```
Jeff-Meadows: I'm not sure `ValueError` is correct here. I'd raise `RuntimeError` or a custom `Exception`

jmoldow: `ValueError` is what the standard library uses for similar purposes:

```
jmoldow Lib ((detached from v3.6.2)) $ git grep -C9 "if .*closed.*[:]" | grep -v asyncio | grep -v pyio | grep -v "test" | grep --color=always "  raise "
_compression.py-14-            raise ValueError("I/O operation on closed file")
_compression.py-18-            raise io.UnsupportedOperation("File not open for reading")
_compression.py-22-            raise io.UnsupportedOperation("File not open for writing")
chunk.py-95-            raise ValueError("I/O operation on closed file")
chunk.py-105-            raise ValueError("I/O operation on closed file")
chunk.py-107-            raise OSError("cannot seek")
chunk.py-113-            raise RuntimeError
chunk.py-119-            raise ValueError("I/O operation on closed file")
chunk.py-129-            raise ValueError("I/O operation on closed file")
chunk.py-153-            raise ValueError("I/O operation on closed file")
idlelib/run.py-339-            raise ValueError("write to closed file")
idlelib/run.py-342-                raise TypeError('must be str, not ' + type(s).__name__)
idlelib/run.py-359-            raise ValueError("read from closed file")
idlelib/run.py-363-            raise TypeError('must be int, not ' + type(size).__name__)
idlelib/run.py-382-            raise ValueError("read from closed file")
idlelib/run.py-386-            raise TypeError('must be int, not ' + type(size).__name__)
pathlib.py-1011-        raise ValueError("I/O operation on closed path")
pathlib.py-1072-            raise ValueError("Unacceptable pattern: {!r}".format(pattern))
pathlib.py-1191-            raise TypeError('data must be str, not %s' %
socket.py-615-            raise ValueError("I/O operation on closed socket.")
socket.py-622-            raise ValueError("I/O operation on closed socket.")
socket.py-629-            raise ValueError("I/O operation on closed socket.")
subprocess.py-188-            raise ValueError("already closed")
tarfile.py-408-                raise CompressionError("unknown compression type %r" % comptype)
tarfile.py-2365-            raise OSError("%s is closed" % self.__class__.__name__)
tarfile.py-2367-            raise OSError("bad operation for mode %r" % self.mode)
tempfile.py-678-            raise ValueError("Cannot enter context with closed file")
urllib/response.py-30-            raise ValueError("I/O operation on closed file")
zipfile.py-985-            raise ValueError('I/O operation on closed file.')
```

jmoldow:
@Jeff-Meadows see my comment in #175. I experimented with adding a `__del__` for `OAuth2`, but I didn't like the behavior, so I removed it.

For the client, we could think about making it into a context manager, which invoke's `auth.closing()`.

PR #225, opened 2017-08-22
Remove unneeded private attribute from client

jmoldow:


boxcla:

PR #226, opened 2017-08-25
Create RequestsSessionNetwork

jmoldow:
This class is the same as the former `DefaultNetwork` class,
except that it accepts an optional `session` constructor
parameter, for clients to specify a session of their choosing
rather than having it constructed automatically.

`DefaultNetwork` is aliased to `RequestsSessionNetwork`, for
backwards-compatibility.

This enables the boxsdk's testing framework to start using the
`betamax` to record real network calls and responses. Although
we currently only add one such test, this is a potential
alternative to mock-box functional tests in the future.

This also allows for dependents of boxsdk to write `betamax`
tests as well.


boxcla:

PR #230, opened 2017-09-12
Add OAuth2.downscope_token() method (fixes #229)

### LINKED ISSUE #229: Implement token downscoping
Add a method to OAuth2 to allow token downscoping based on https://developer.box.com/reference#token-exchange

anthonywee:
This allows the caller to obtain a token with its
permissions reduced to the provided scope(s) and optionally for the
provided file or folder.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -181,7 +181,7 @@ class OAuth2(object):
             data['box_device_id'] = self._box_device_id
         if self._box_device_name:
             data['box_device_name'] = self._box_device_name
-        return self.send_token_request_and_store_tokens(data, access_token=None)
+        return self.send_token_request(data, access_token=None)
 
     def _refresh(self, access_token):
         data = {
@@ -195,7 +195,7 @@ class OAuth2(object):
         if self._box_device_name:
             data['box_device_name'] = self._box_device_name
 
-        return self.send_token_request_and_store_tokens(data, access_token)
+        return self.send_token_request(data, access_token)
 
     def _get_tokens(self):
         """
@@ -294,7 +294,7 @@ class OAuth2(object):
         """
         self._access_token, self._refresh_token = access_token, refresh_token
 
-    def _send_token_request(self, data, access_token, expect_refresh_token=True):
+    def _send_token_request_without_storing_tokens(self, data, access_token, expect_refresh_token=True):
         """
         Send the request to acquire or refresh an access token.
 
@@ -334,7 +334,7 @@ class OAuth2(object):
 
         return access_token, refresh_token
 
-    def send_token_request_and_store_tokens(self, data, access_token, expect_refresh_token=True):
+    def send_token_request(self, data, access_token, expect_refresh_token=True):
         """
         Send the request to acquire or refresh an access token, and store the tokens.
 
@@ -351,7 +351,7 @@ class OAuth2(object):
         :rtype:
             (`unicode`, `unicode`)
         """
-        access_token, refresh_token = self._send_token_request(data, access_token, expect_refresh_token)
+        access_token, refresh_token = self._send_token_request_without_storing_tokens(data, access_token, expect_refresh_token)
         self._store_tokens(access_token, refresh_token)
         return self._access_token, self._refresh_token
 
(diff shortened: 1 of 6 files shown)


boxcla:

boxsdk/auth/oauth2.py
```diff
@@ -343,6 +354,57 @@ def revoke(self):
                 raise BoxOAuthException(network_response.status_code, network_response.content, url, 'POST')
             self._store_tokens(None, None)
 
+    def downscope_token(self, scopes, item=None, additional_data=None):
+        """
+        Get a downscoped token for the provided file or folder with the provided scopes
```
Jeff-Meadows: Document what happens when item is None

boxsdk/auth/oauth2.py
```diff
@@ -343,6 +354,57 @@ def revoke(self):
...
+
+        :param scope:
+            The scope(s) to apply to the resulting token
+        :type scopes:
+            `list` of :class:`TokenScope`
```
Jeff-Meadows: `Iterable` instead of `list`?

boxsdk/auth/oauth2.py
```diff
@@ -14,6 +14,17 @@
 from boxsdk.network.default_network import DefaultNetwork
 from boxsdk.config import API
 from boxsdk.exception import BoxOAuthException
+from boxsdk.util.text_enum import TextEnum
+
+
+class TokenScope(TextEnum):
```
jmoldow: Docstring, and link to documentation that defines these.

boxsdk/auth/oauth2.py
```diff
@@ -343,6 +354,57 @@ def revoke(self):
...
+            `dict`
+        :return:
+            The downscoped token
+        :rtype:
+            `unicode`
+        """
+        self._check_closed()
+        with self._refresh_lock:
+            self._check_closed()
+            url = '{base_auth_url}/token'.format(base_auth_url=API.OAUTH2_API_URL)
```
jmoldow: Rather than re-implementing most of `send_token_request`, can we refactor part of the definition of `send_token_request` into a new (private) method, which makes the API call but doesn't call `_store_tokens`? And then this method and `send_token_request` can both call that method.

In the future that could even be a public method, but it should probably be private to start with.

Jeff-Meadows: I was actually thinking we should call the new method `_send_token_request` and rename `send_token_request` -> `send_token_request_and_store_tokens`

boxsdk/auth/oauth2.py
anthonywee: Done

boxsdk/auth/oauth2.py
anthonywee: done

boxsdk/auth/oauth2.py
anthonywee: Good idea, done

boxsdk/auth/oauth2.py
anthonywee: done

boxsdk/auth/oauth2.py
```diff
@@ -166,7 +181,7 @@ def authenticate(self, auth_code):
             data['box_device_id'] = self._box_device_id
         if self._box_device_name:
             data['box_device_name'] = self._box_device_name
-        return self.send_token_request(data, access_token=None)
+        return self.send_token_request_and_store_tokens(data, access_token=None)
```
Jeff-Meadows: Sorry, I know I suggested this rename, but then I realized it's an unnecessary breaking change. Can you rename this back, and call the new method something else like `_send_token_request_without_storing_tokens`?

jmoldow: +1

boxsdk/auth/oauth2.py
```diff
@@ -343,6 +379,46 @@ def revoke(self):
...
+        :param additional_data:
+            (Optional) Key value pairs which can be used to add/update the default data values in the request.
+        :type additional_data:
+            `dict`
+        :return:
+            The downscoped token
+        :rtype:
+            `unicode`
+        """
+        self._check_closed()
```
jmoldow: Change this to use `_send_token_request_without_storing_tokens`.

boxsdk/auth/oauth2.py
anthonywee: Done

PR #232, opened 2017-09-13
Bug fix: impossible to use RedisManagedJWTAuth (closes #231)

### LINKED ISSUE #231: Bug when using RedisManagedJWTAuth
If I try to use RedisManagedJWTAuth instead of JWTAuth, I got this error:
`__init__() got an unexpected keyword argument 'refresh_lock'`

Seems this is because RedisManagedOAuth2Mixin calls parent class (JWTAuth) with extra argument (`refresh_lock=`), which is absent in that class, but present in grand parent class (OAuth2).

alextop:
Fixes `__init__()` called from RedisManagedOAuth2Mixin with extra argument (`refresh_lock=`), which is absent in parent class.

### DIFF
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -34,6 +34,7 @@ class JWTAuth(OAuth2):
             access_token=None,
             network_layer=None,
             jwt_algorithm='RS256',
+            refresh_lock=None,
     ):
         """
         :param client_id:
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @alextop, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

jmoldow:
A more flexible solution is in #233. But thanks for the PR and the bug report!

PR #233, opened 2017-09-13
Add kwargs to JWTAuth constructor

### LINKED ISSUE #231: Bug when using RedisManagedJWTAuth

jmoldow:
Fixes #231.


boxcla:

PR #235, opened 2017-09-13
View path collabs

potrebic:
Add support for view_path collabs.

Adding ability to set the `can_view_path` body param when constructing new folder collaborations.

ref: https://developer.box.com/v2.0/reference#add-a-collaboration


boxcla:

PR #236, opened 2017-09-13
Add support for view_path collabs

potrebic:
Adding ability to set the `can_view_path` body param when constructing
new folder collaborations.

ref: https://developer.box.com/v2.0/reference#add-a-collaboration

### DIFF
--- a/boxsdk/object/folder.py
+++ b/boxsdk/object/folder.py
@@ -408,7 +408,8 @@ class Folder(Item):
         :type notify:
             `bool`
         :param can_view_path:
-            Whether view path collaboration feature is enabled or not
+            Whether view path collaboration feature is enabled or not. Note - only
+            folder owners can create collaborations with can_view_path.
         :type can_view_path:
             `bool`
         :return:
--- a/boxsdk/object/folder.py
+++ b/boxsdk/object/folder.py
@@ -392,7 +392,7 @@ class Folder(Item):
         return self.update_info(data=data)
 
     @api_call
-    def add_collaborator(self, collaborator, role, notify=False):
+    def add_collaborator(self, collaborator, role, notify=False, can_view_path=False):
         """Add a collaborator to the folder
 
         :param collaborator:
@@ -407,6 +407,10 @@ class Folder(Item):
             Whether to send a notification email to the collaborator
         :type notify:
             `bool`
+        :param can_view_path:
+            Whether view path collaboration feature is enabled or not
+        :type can_view_path:
+            `bool`
         :return:
             The new collaboration
         :rtype:
@@ -418,13 +422,16 @@ class Folder(Item):
         access_key, access_value = collaborator_helper.access
         accessible_by = {
             access_key: access_value,
-            'type': collaborator_helper.type
+            'type': collaborator_helper.type,
         }
-        data = json.dumps({
+        body_params = {
             'item': item,
             'accessible_by': accessible_by,
             'role': role,
-        })
+        }
+        if can_view_path:
+            body_params['can_view_path'] = True
+        data = json.dumps(body_params)
         params = {'notify': notify}
         box_response = self._session.post(url, expect_json_response=True, data=data, params=params)
         collaboration_response = box_response.json()
--- a/test/unit/object/test_folder.py
+++ b/test/unit/object/test_folder.py
@@ -67,47 +67,37 @@ def mock_items_response(mock_items):
     return get_response
 
 
-def _assert_collaborator_added(test_folder, collaborator, mock_box_session, mock_collab_response, notify, role, data):
+def _assert_collaborator_added(test_folder, collaborator, mock_box_session, mock_collab_response, notify, role, can_view_path, data):
     mock_box_session.post.return_value = mock_collab_response
-    collaboration = test_folder.add_collaborator(collaborator, role, notify)
+    collaboration = test_folder.add_collaborator(collaborator, role, notify, can_view_path)
     assert isinstance(collaboration, Collaboration)
     expected_url = API.BASE_API_URL + '/collaborations'
     params = {'notify': notify}
     mock_box_session.post.assert_called_once_with(expected_url, expect_json_response=True, data=data, params=params)
 
 
+@pytest.mark.parametrize('accessible_by', ['user', 'group', 'email'])
 @pytest.mark.parametrize('notify', [True, False])
 @pytest.mark.parametrize('role', iter(CollaborationRole))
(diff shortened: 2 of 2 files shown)


boxcla:

test/unit/object/test_folder.py
```diff
@@ -67,47 +67,37 @@ def get_response(limit, offset):
...
-def test_add_group_collaborator(test_folder, mock_group, mock_box_session, mock_collab_response, notify, role):
-    data = json.dumps({
-        'item': {'id': test_folder.object_id, 'type': 'folder'},
-        'accessible_by': {'id': mock_group.object_id, 'type': 'group'},
-        'role': role,
-    })
-    _assert_collaborator_added(test_folder, mock_group, mock_box_session, mock_collab_response, notify, role, data)
+@pytest.mark.parametrize('can_view_path', [True, False])
+def test_add_collaborator(test_folder, mock_user, mock_group, mock_box_session, mock_collab_response, accessible_by, notify, role, can_view_path):
+    accessible_dict = {
```
potrebic: I made this change to get rid of the duplication in the 3 previous tests
> test_add_user_collaborator
> test_add_group_collaborator
> test_add_email_collaborator

Which would have gotten worse with the new `can_view_path` param.

boxsdk/object/folder.py
```diff
@@ -392,7 +392,7 @@ def update_sync_state(self, sync_state):
         return self.update_info(data=data)
 
     @api_call
-    def add_collaborator(self, collaborator, role, notify=False):
+    def add_collaborator(self, collaborator, role, notify=False, can_view_path=False):
```
jmoldow: Two things I wonder about this:

1) Can we safely put this before `notify`? It'd be nice to put all of the object params before the `notify` param, since the latter isn't really related to business logic. The question is, how many people use `notify`, and how many of those use it as a positional argument? Even if we're worried about that, we can still make it as a breaking change, since we're in the transition to 2.0.0.

2) Rather than stacking a bunch of optional params at the end (it's only `can_view_path` for now, but who knows, in the future there could be others), I wonder if we should instead expose a generic parameter for extra body params. Something like `(additional | extra | optional)_attributes`. And then callers could do `additional_attributes={'can_view_path': True, 'foo': 'bar'}`

Thoughts @Jeff-Meadows

boxsdk/object/folder.py
```diff
@@ -407,6 +407,10 @@ def add_collaborator(self, collaborator, role, notify=False):
             Whether to send a notification email to the collaborator
         :type notify:
             `bool`
+        :param can_view_path:
+            Whether view path collaboration feature is enabled or not
```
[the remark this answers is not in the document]
potrebic: will do

test/unit/object/test_folder.py
```diff
@@ -67,47 +67,37 @@ def get_response(limit, offset):
...
+    }
 
+    invitee, mock_accessible_by = accessible_dict[accessible_by]
 
-@pytest.mark.parametrize('notify', [True, False])
-@pytest.mark.parametrize('role', iter(CollaborationRole))
-def test_add_email_collaborator(test_folder, mock_box_session, mock_collab_response, notify, role):
-    email_address = 'foo@example.com'
-    data = json.dumps({
+    body_parems = {
```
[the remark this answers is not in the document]
potrebic: Will do

test/unit/object/test_folder.py
```diff
@@ -67,47 +67,37 @@ def get_response(limit, offset):
...
-        'accessible_by': {'id': mock_group.object_id, 'type': 'group'},
-        'role': role,
-    })
-    _assert_collaborator_added(test_folder, mock_group, mock_box_session, mock_collab_response, notify, role, data)
+@pytest.mark.parametrize('can_view_path', [True, False])
+def test_add_collaborator(test_folder, mock_user, mock_group, mock_box_session, mock_collab_response, accessible_by, notify, role, can_view_path):
+    accessible_dict = {
+        'user': (mock_user, {'id': mock_user.object_id, 'type': 'user'}),
+        'group': (mock_group, {'id': mock_group.object_id, 'type': 'group'}),
+        'email': ('foo@example.com', {'login': 'foo@example.com', 'type': 'user'}),
```
jmoldow: This is fine, but you also could've done something like this:

```python
@pytest.mark.parametrize(
    ('accessible_by', 'mock_accessible_by'),
    [
        ('user', {'id': mock_user.object_id, 'type': 'user'}),
        ...
    ],
)
```

(you would still need the map for extracting the invitee, since those are other fixtures)

or something like this:

```python
@pytest.fixture
def invitee(accessible_by, mock_user, mock_group):
    return {'user': mock_user, 'group': mock_group, 'email': 'foo@example.com'}[accessible_by]

@pytest.fixture
def mock_accessible_by(accessible_by):
    return {
        'user': {'id': mock_user.object_id, 'type': 'user'},
        ...
    }[accessible_by]

@pytest.mark.parametrize('accessible_by', ['user', 'group', 'email'])
def test_add_collaborator(mock_accessible_by, invitee):
    ...
```

Either of these approaches would keep most, or all, of the fixture data out of the body of the test.

boxsdk/object/folder.py
potrebic: 1. I'm not sure why `notify` isn't business logic. The SDK doesn't get to specify what is or isn't business logic. Sending or not emails could very much be business logic???

That said I'm fine placing this param anywhere.

2. I thought about that option and then went to the Docs and saw that can_view_path was the only unsupported param.

But thinking long term / broadly... having some sort of policy/pattern in the SDK's API that maps nicely to the actual V2 API would be nice. Something like::
* required params have explicit named-params.
* optionals params are supported via the optional dicts.

Jordan - are you suggesting just body-params would have a dict parameter? What about query params?

Looking at https://developer.box.com/v2.0/reference#add-a-collaboration
* query param `fields`. its optional. The box-sdk doesn't support it.
* query param `notify` the docs don't list it's default. But I *guess* it's optional.
* body param `accessible_by`. It's not labeled as Required. Is that correct
* body param `can_view_path`. Optional, so it could be cover by a dict param

Thus the current boxSDK API "breaks" the model of optional being supported via dicts. So what then... it's just random. Some params are explicit. Others are supported via dict? Maybe that is OK. Maybe not. Or maybe in the move to 2.0 would implement a nice pattern like this? That would mean removing the `notify` param and having a "option_query_params" dict parameter.

potrebic: Jeff and I discussed this and agreed that the discussion / implementation of a pattern for supporting all current and future query-params and body params is beyond the scope of this change. And that this change as-is is ok.

PR #237, opened 2017-09-16
Change downscope_token() to return a TokenResponse object

anthonywee:
- Add a TokenResponse object, which is a subclass of BaseAPIJSONObject.
- OAuth2.downscope_token() now returns a TokenResponse object, which
  allows users to get other fields of the response, such as
  'expires_in'.

### DIFF
--- a/boxsdk/auth/oauth2.py
+++ b/boxsdk/auth/oauth2.py
@@ -11,9 +11,10 @@ import sys
 import six
 from six.moves.urllib.parse import urlencode, urlunsplit  # pylint:disable=import-error,no-name-in-module
 
-from boxsdk.network.default_network import DefaultNetwork
 from boxsdk.config import API
 from boxsdk.exception import BoxOAuthException
+from boxsdk.network.default_network import DefaultNetwork
+from boxsdk.object.base_api_json_object import BaseAPIJSONObject
 from boxsdk.util.text_enum import TextEnum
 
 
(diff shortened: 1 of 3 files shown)


boxcla:

boxsdk/auth/oauth2.py
```diff
@@ -416,8 +422,9 @@ def downscope_token(self, scopes, item=None, additional_data=None):
             if additional_data:
                 data.update(additional_data)
 
-            access_token, _ = self._send_token_request_without_storing_tokens(data, access_token, expect_refresh_token=False)
-        return access_token
+            token_response = self.execute_token_request(data, access_token, expect_refresh_token=False)
```
jmoldow: Just return from here.

boxsdk/auth/oauth2.py
```diff
@@ -294,7 +300,7 @@ def _update_current_tokens(self, access_token, refresh_token):
         """
         self._access_token, self._refresh_token = access_token, refresh_token
 
-    def _send_token_request_without_storing_tokens(self, data, access_token, expect_refresh_token=True):
+    def execute_token_request(self, data, access_token, expect_refresh_token=True):
```
jmoldow: I think we should keep this private for now.

Not that I think it should never be public. But until we know that we definitely need this in the public API, I don't think we should clutter the API and add more backwards-compatibility restrictions for ourselves.

boxsdk/auth/oauth2.py
anthonywee: done

boxsdk/auth/oauth2.py
anthonywee: sure, done

PR #240, opened 2017-09-18
Fix some issues with six dependency

### LINKED ISSUE #228: `six.raise_from` introduced in 1.9.0 but installation only requires six>=1.4.0
Some recent commits began using `raise_from` from `six`, but didn't bump the requirement to `six>=1.9.0`.

### LINKED ISSUE #239: boxsdk 2 not compatible with six==1.11.0
Hello,

a new version of the six module has been released, boxsdk is not compatible.

to replicate the problem:

pip install six==1.11.0 boxsdk==2.0.0a8
python -c "import boxsdk"

results in:
```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/boxsdk/__init__.py", line 6, in <module>
    from .client import *  # pylint:disable=wildcard-import,redefined-builtin
  File "/usr/local/lib/python2.7/site-packages/boxsdk/client/__init__.py", line 5, in <module>
    from .client import Client
  File "/usr/local/lib/python2.7/site-packages/boxsdk/client/client.py", line 12, in <module>
    from ..object.events import Events
  File "/usr/local/lib/python2.7/site-packages/boxsdk/object/events.py", line 15, in <module>
    class EventsStreamType(with_metaclass(ExtendableEnumMeta, TextEnum)):
  File "/usr/local/lib/python2.7/site-packages/boxsdk/util/compat.py", line 77, in with_metaclass
    return type.__new__(TemporaryMetaSubclass, str('temporary_class'), bases, {})
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

I reported the bug in six and they have responded, suggesting a change in box:
https://github.com/benjaminp/six/issues/210#issuecomment-330206799

our workaround for the moment is to pin six to version 1.10, i link to this here for posterity.

cheers

jmoldow:
Add a fallback for `six.raise_from`, which isn't available in
six 1.4.0. It isn't available until six 1.9.0. We could also
have raised the lower bound for the six requirement, but this is
an easy way to allow clients to keep using their existing
versions of six.

Fix support for the latest version of six, 1.11.0. That release
changed the temporary metaclass returned from
`with_metaclass()`, such that it directly inherits from `type`,
instead of inheriting from the target metaclass [1]. We depended
on this detail, and the change caused

.. code-block:: python

    TypeError('metaclass conflict: ...')

to be raised when defining a class with `with_metaclass()`. We
fix this by manually selecting the most derived metaclass, and
including it in our temporary metaclass.

Also, `__prepare__` is now defined on the temporary metaclass,
in six 1.11.0 [2]. This allows us to skip our own definition of that
method, when using six>=1.11.0.

Fixes #228.

Fixes #239.

[1] <https://github.com/benjaminp/six/pull/191>
[2] <https://github.com/benjaminp/six/pull/178>

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -134,6 +134,7 @@ Release History
   when it tried to refresh any ``JWTAuth`` object.
 - Fixed an exception that was being raised from ``ExtendableEnumMeta.__dir__()``.
 - CPython 3.6 support.
+- Increased required minimum version of six to 1.9.0.
 
 1.5.3 (2016-05-26)
 ++++++++++++++++++
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -10,11 +10,11 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
 import jwt
-from six import binary_type, string_types, raise_from, text_type
+from six import binary_type, string_types, text_type
 
 from .oauth2 import OAuth2
 from ..object.user import User
-from ..util.compat import NoneType, total_seconds
+from ..util.compat import NoneType, raise_from, total_seconds
 
 
 class JWTAuth(OAuth2):
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -10,11 +10,11 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
 import jwt
-from six import binary_type, string_types, text_type
+from six import binary_type, string_types, raise_from, text_type
 
 from .oauth2 import OAuth2
 from ..object.user import User
-from ..util.compat import NoneType, raise_from, total_seconds
+from ..util.compat import NoneType, total_seconds
 
 
 class JWTAuth(OAuth2):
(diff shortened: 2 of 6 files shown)


boxcla:

Jeff-Meadows:
If we're going to reimplement `raise_from` in case it's not defined in the currently installed version of `six`, shouldn't we define it correctly (ie using the `raise from` syntax if we're in `PY3`)?

jmoldow:
@Jeff-Meadows I didn't want to simply redefine it, because then we'd have code that is identical to what appears in six, and I don't want to get into any licensing grey areas.

I think the way it is is okay. In Python 2, there is no difference. In Python 3, the only difference is the `__cause__` attribute, which very few clients will look at. Practically, it only really matters for stacktraces.

My personal preference would be:

1. Keep as-is, or bump the minimum required version of six.
2. Remove use of `raise_from`.
3. Fully redefine `raise_from` to match six.

I'd be okay with not using `raise_from`, since it's more of a nice optimization, rather than something that we must do.

But since six 1.9.0 is over two years old, and since we're doing a major version bump, I guess I'd be fine changing this to increase the minimum version, if we don't want to do weird things like redefine a partially-implemented `raise_from`.

Jeff-Meadows:
@jmoldow I'd rather not reimplement than partially reimplement. I think bumping the minimum to `1.9.0` is perfectly fine (I'd also be fine bumping to `1.11.0` in the `2.0` branch and removing the additional `with_metaclass`  code)

jmoldow:
Bumped the minimum to 1.9.0.

jmoldow:
I removed the entries from requirements.txt, since I don't believe there is anything gained from duplicating everything.

PR #245, opened 2017-09-28
Add methods to configure JWT auth from a config file.

### LINKED ISSUE #221: Add support to load app config file downloaded from dev console
Add support to jwt_auth.py to load a box_config.json file downloaded from the developer console. This would make the python SDK consistent with the node an java sdk's.

node: https://github.com/box/box-node-sdk/blob/3dcfe69ed1354ed85b54eb4b1d4e179c999f4f99/lib/box-node-sdk.js#L126

java: 
https://github.com/box/box-java-sdk/blob/master/src/main/java/com/box/sdk/BoxConfig.java
https://github.com/box/box-java-sdk/blob/da3332d77c09f1c75afe93184b2a02bb00e84dd8/src/main/java/com/box/sdk/BoxDeveloperEditionAPIConnection.java#L189

Jeff-Meadows:
Fixes #221.

The Box developer console allows a developer to download a
configuration file that specifies how to configure a Box
application.

This commit adds code to parse these config files and apply
the configuration to a JWTAuth instance.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -105,6 +105,8 @@ Release History
 - Added a ``downscope_token()`` method to the ``OAuth2`` class. This generates a token that
   has its permissions reduced to the provided scopes and for the optionally provided 
   ``File`` or ``Folder``.
+- Added methods for configuring `JWTAuth` from config file: `JWTAuth.from_config_file` and
+  `JWTAuth.from_config_dictionary`.
 
 **Other**
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -105,8 +105,8 @@ Release History
 - Added a ``downscope_token()`` method to the ``OAuth2`` class. This generates a token that
   has its permissions reduced to the provided scopes and for the optionally provided 
   ``File`` or ``Folder``.
-- Added methods for configuring `JWTAuth` from config file: `JWTAuth.from_config_file` and
-  `JWTAuth.from_config_dictionary`.
+- Added methods for configuring ``JWTAuth`` from config file: ``JWTAuth.from_config_file`` and
+  ``JWTAuth.from_config_dictionary``.
 
 **Other**
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -105,8 +105,8 @@ Release History
 - Added a ``downscope_token()`` method to the ``OAuth2`` class. This generates a token that
   has its permissions reduced to the provided scopes and for the optionally provided 
   ``File`` or ``Folder``.
-- Added methods for configuring ``JWTAuth`` from config file: ``JWTAuth.from_config_file`` and
-  ``JWTAuth.from_config_dictionary``.
+- Added methods for configuring ``JWTAuth`` from config file: ``JWTAuth.from_settings_file`` and
+  ``JWTAuth.from_settings_dictionary``.
 
 **Other**
 
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -363,40 +363,40 @@ class JWTAuth(OAuth2):
         return passphrase
 
     @classmethod
-    def from_settings_dictionary(cls, config_dictionary, **kwargs):
+    def from_settings_dictionary(cls, settings_dictionary, **kwargs):
         """
         Create an auth instance as defined by the given settings dictionary.
 
         The dictionary should have the structure of the JSON file downloaded from the Box Developer Console.
 
-        :param config_dictionary:       Dictionary containing settings for configuring app auth.
-        :type config_dictionary:        `dict`
+        :param settings_dictionary:       Dictionary containing settings for configuring app auth.
+        :type settings_dictionary:        `dict`
         :return:                        Auth instance configured as specified by the config dictionary.
         :rtype:                         :class:`JWTAuth`
         """
-        if 'boxAppSettings' not in config_dictionary:
+        if 'boxAppSettings' not in settings_dictionary:
             raise ValueError('boxAppSettings not present in configuration')
         return cls(
-            client_id=config_dictionary['boxAppSettings']['clientID'],
-            client_secret=config_dictionary['boxAppSettings']['clientSecret'],
-            enterprise_id=config_dictionary.get('enterpriseID', None),
-            jwt_key_id=config_dictionary['boxAppSettings']['appAuth'].get('publicKeyID', None),
-            rsa_private_key_data=config_dictionary['boxAppSettings']['appAuth'].get('privateKey', None),
-            rsa_private_key_passphrase=config_dictionary['boxAppSettings']['appAuth'].get('passphrase', None),
+            client_id=settings_dictionary['boxAppSettings']['clientID'],
+            client_secret=settings_dictionary['boxAppSettings']['clientSecret'],
+            enterprise_id=settings_dictionary.get('enterpriseID', None),
+            jwt_key_id=settings_dictionary['boxAppSettings']['appAuth'].get('publicKeyID', None),
+            rsa_private_key_data=settings_dictionary['boxAppSettings']['appAuth'].get('privateKey', None),
+            rsa_private_key_passphrase=settings_dictionary['boxAppSettings']['appAuth'].get('passphrase', None),
             **kwargs
         )
 
     @classmethod
-    def from_config_file(cls, config_file_sys_path, **kwargs):
+    def from_settings_file(cls, settings_file_sys_path, **kwargs):
         """
         Create an auth instance as defined by a JSON file downloaded from the Box Developer Console.
         See https://developer.box.com/v2.0/docs/authentication-with-jwt for more information.
 
-        :param config_file_sys_path:    Path to the JSON file containing the configuration.
-        :type config_file_sys_path:     `unicode`
+        :param settings_file_sys_path:    Path to the JSON file containing the configuration.
+        :type settings_file_sys_path:     `unicode`
         :return:                        Auth instance configured as specified by the JSON file.
         :rtype:                         :class:`JWTAuth`
         """
-        with open(config_file_sys_path) as config_file:
+        with open(settings_file_sys_path) as config_file:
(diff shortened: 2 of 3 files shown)


boxcla:

HISTORY.rst
```diff
@@ -105,6 +105,8 @@ Release History
 - Added a ``downscope_token()`` method to the ``OAuth2`` class. This generates a token that
   has its permissions reduced to the provided scopes and for the optionally provided 
   ``File`` or ``Folder``.
+- Added methods for configuring `JWTAuth` from config file: `JWTAuth.from_config_file` and
```
jmoldow: Needs double-backticks to render properly.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,51 @@ def _normalize_rsa_private_key_passphrase(passphrase):
                 .format(passphrase.__class__.__name__)
             )
         return passphrase
+
+    @classmethod
+    def from_config_dictionary(cls, config_dictionary):
```
jmoldow: Nitpick, but I somewhat lean towards "settings" over "config". Since the JSON dictionary has a top-level name "boxAppSettings", and the admin console section is named "App Settings".

I'm not totally opposed to "config" though.

jmoldow: This method, and also `from_config_file`, should accept `**kwargs`.

We still want users of these constructors to be able to, e.g. specify their own network layer, or specify a different algorithm.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,51 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+        """
+        Create an auth instance as defined by the given configuration dictionary.
+
+        The dictionary should have the following structure:
```
jmoldow: Rather than documenting the structure here, maybe just point them to the place to find this in the developer app console.

I don't think we want users to think that we're encouraging them to manually construct these JSON objects. At that point, they're better off calling the constructor manually.

Also, link to any documentation that there might be about this feature.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,51 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+        :return:                        Auth instance configured as specified by the config dictionary.
+        :rtype:                         :class:`JWTAuth`
+        """
+        if 'boxAppSettings' not in config_dictionary:
+            raise ValueError('boxAppSettings not present in configuration')
+        return cls(
+            client_id=config_dictionary['boxAppSettings']['clientID'],
+            client_secret=config_dictionary['boxAppSettings']['clientSecret'],
+            enterprise_id=config_dictionary.get('enterpriseID'),
+            jwt_key_id=config_dictionary['boxAppSettings']['appAuth']['publicKeyID'],
```
jmoldow: I think all of the `appAuth` settings should use `['appAuth'].get('publicKeyID') or None` instead of `['appAuth']['publicKeyID']`. Three reasons for that:

1) If they forget to enter the correct values, the constructor will treat it as if the user didn't pass anything, instead of 0-length data (since the developer console defaults to showing these as empty strings).
2) In particular, it might be the case (though I'm not sure) that cryptography would treat an empty string passphrase differently from `None`.
3) This would allow the user to skip having to define `privateKey`, if they instead wanted to pass a system path (`rsa_private_key_file_sys_path`) via the keyword arguments I suggested.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,51 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+            rsa_private_key_data=config_dictionary['boxAppSettings']['appAuth']['privateKey'],
+            rsa_private_key_passphrase=config_dictionary['boxAppSettings']['appAuth']['passphrase'],
+        )
+
+    @classmethod
+    def from_config_file(cls, config_file_sys_path):
+        """
+        Create an auth instance as defined by a JSON file downloaded from the Box Developer Console.
```
jmoldow: If there is any documentation about this, link to it.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,51 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+        :type config_dictionary:        `dict`
+        :return:                        Auth instance configured as specified by the config dictionary.
+        :rtype:                         :class:`JWTAuth`
+        """
+        if 'boxAppSettings' not in config_dictionary:
+            raise ValueError('boxAppSettings not present in configuration')
+        return cls(
+            client_id=config_dictionary['boxAppSettings']['clientID'],
+            client_secret=config_dictionary['boxAppSettings']['clientSecret'],
+            enterprise_id=config_dictionary.get('enterpriseID'),
```
jmoldow: Should this be `config_dictionary.get('enterpriseID') or None`?

test/unit/auth/test_jwt_auth.py
```diff
@@ -364,3 +364,99 @@ def test_refresh_instance_sends_post_request_with_correct_params(jwt_auth_init_a
...
+):
+    return r"""
+{{
+  "boxAppSettings": {{
+    "clientID": "{client_id}",
+    "clientSecret": "{client_secret}",
+    "appAuth": {{
+      "publicKeyID": "{jwt_key_id}",
+      "privateKey": "{private_key}",
+      "passphrase": "{passphrase}"
```
jmoldow: This isn't quite right, because:

If the `rsa_passphrase` fixture is ever `None`, it will get written as `"None"` instead of `""` or `null`.

Try `passphrase=(rsa_passphrase or '')` in the `.format()` call.

Similarly for `fake_enterprise_id`.

test/unit/auth/test_jwt_auth.py
```diff
@@ -364,3 +364,99 @@ def test_refresh_instance_sends_post_request_with_correct_params(jwt_auth_init_a
     enterprise_id = 'fake_enterprise_id'
     with jwt_auth_init_and_auth_mocks(enterprise_id, 'enterprise', enterprise_id=enterprise_id) as oauth:
         oauth.refresh(None)
+
+
+@pytest.fixture()
+def jwt_subclass_that_just_stores_params():
+    class StoreParamJWTAuth(JWTAuth):
+        def __init__(self, **kwargs):
+            self.kwargs = kwargs
```
jmoldow: Can you call `super()` here?

If you don't, then this test is incomplete. If we were to ever change our input params (e.g. say we rename a param), this test would continue to pass. Whereas if we call `super()`, then the super call would fail in such a scenario.

Since you are using `rsa_private_key_bytes`, that should work without any mocking needed.

test/unit/auth/test_jwt_auth.py
```diff
@@ -364,3 +364,99 @@ def test_refresh_instance_sends_post_request_with_correct_params(jwt_auth_init_a
...
+        rsa_private_key_bytes,
+        rsa_passphrase,
+):
+    def _assert_jwt_kwargs_expected(jwt_auth):
+        assert jwt_auth.kwargs['client_id'] == fake_client_id
+        assert jwt_auth.kwargs['client_secret'] == fake_client_secret
+        assert jwt_auth.kwargs['enterprise_id'] == fake_enterprise_id
+        assert jwt_auth.kwargs['jwt_key_id'] == jwt_key_id
+        assert jwt_auth.kwargs['rsa_private_key_data'] == rsa_private_key_bytes.replace("\n", '')
+        assert jwt_auth.kwargs['rsa_private_key_passphrase'] == text_type(rsa_passphrase)
```
jmoldow: This should just be `rsa_passphrase`. I think `text_type()` is masking the problem I mentioned above.

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,42 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+            client_secret=config_dictionary['boxAppSettings']['clientSecret'],
+            enterprise_id=config_dictionary.get('enterpriseID', None),
+            jwt_key_id=config_dictionary['boxAppSettings']['appAuth'].get('publicKeyID', None),
+            rsa_private_key_data=config_dictionary['boxAppSettings']['appAuth'].get('privateKey', None),
+            rsa_private_key_passphrase=config_dictionary['boxAppSettings']['appAuth'].get('passphrase', None),
+            **kwargs
+        )
+
+    @classmethod
+    def from_config_file(cls, config_file_sys_path, **kwargs):
```
jmoldow: Named now mismatch. Can you make this one `from_settings_*` as well?

And can you name the params for both methonds `settings_*`?

boxsdk/auth/jwt_auth.py
```diff
@@ -360,3 +361,42 @@ def _normalize_rsa_private_key_passphrase(passphrase):
...
+        :type config_dictionary:        `dict`
+        :return:                        Auth instance configured as specified by the config dictionary.
+        :rtype:                         :class:`JWTAuth`
+        """
+        if 'boxAppSettings' not in config_dictionary:
+            raise ValueError('boxAppSettings not present in configuration')
+        return cls(
+            client_id=config_dictionary['boxAppSettings']['clientID'],
+            client_secret=config_dictionary['boxAppSettings']['clientSecret'],
+            enterprise_id=config_dictionary.get('enterpriseID', None),
```
jmoldow: This is different than `.get('enterpriseID') or None`. What you have here will return an empty string, if an empty string is given. What I had proposed will return `None`, if an empty string is given.

I know that our functions _sometimes_ treat an empty string the same as `None`, but I think it'd be safer and more correct to pass `None`.

Similarly below.

Jeff-Meadows: The SDK right now does very little sanity checking. I don't think the empty string is a valid rsa passphrase (actually, I have no idea), and I also don't think it's a valid enterprise ID. But I'm not sure we need to special-case these by turning empty string into None.

PR #246, opened 2017-10-03
Add network_response to BoxOAuthException

anthonywee:


boxcla:

anthonywee:
This is a PR to compare 2 options for allowing for retries of a failed downscope_token() request. The other PR is #247

test/unit/test_exception.py
```diff
@@ -47,14 +47,17 @@ def test_box_oauth_exception():
...
         url=url,
         method=method,
+        network_response=network_response,
     )
     assert str(box_exception) == '''
 Message: {0}
 Status: {1}
 URL: {2}
 Method: {3}'''.format(message, status, url, method)
+    assert box_exception.network_response == network_response
```
jmoldow: ` is ` would be more appropriate than ` == ` here.

boxsdk/exception.py
```diff
@@ -193,12 +193,28 @@ def __init__(self, status, message=None, url=None, method=None):
...
         self._url = url
         self._method = method
+        self._network_response = network_response
+
+    @property
+    def network_response(self):
+        """The response for the request.
+
+        :return network_response:
+            The network response for the request.
```
jmoldow: This is exactly the same as what is said one line above. I don't think you need `:return:` for this.

jmoldow:
Can you also add `headers`, and include `headers` in the string representation?

jmoldow:
Please update the `HISTORY` file.

anthonywee:
For `headers`, do you mean the response headers or the request headers? For the response headers, that should be accessible from the `network_response`. Would we expose that as a convenience?

jmoldow:
I guess it doesn't need to be a separate param or a separate property. But can you put it into the message?

PR #249, opened 2017-10-18
Automatically retry JWT auth API calls that fail because of the exp c…

Jeff-Meadows:
…laim.

JWT auth requires an expiration time, but legitimate authorization
requests can be rejected if the system times differ between the
request originator and the Box servers.

This commit adds logic to detect this situation and automatically retry.
The Box servers include a ``Date`` header when the auth request is
rejected due to the exp claim, so we use that value to construct the
new claim.

### DIFF
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -249,7 +249,7 @@ class JWTAuth(OAuth2):
                 error_description = json_response['error_description']
                 box_date_header = error_response.headers['date']
                 box_datetime = datetime.strptime(box_date_header, '%a, %d %b %Y %H:%M:%S %Z')
-            except:  # pylint:disable=broad-except
+            except Exception:  # pylint:disable=broad-except
                 pass
             else:
                 if status_code == 400 and 'exp' in error_description:
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -13,6 +13,7 @@ from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
 import jwt
 from six import binary_type, string_types, raise_from, text_type
 
+from ..exception import BoxOAuthException
 from .oauth2 import OAuth2
 from ..object.user import User
 from ..util.compat import NoneType, total_seconds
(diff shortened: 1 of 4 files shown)


boxcla:

boxsdk/auth/jwt_auth.py
```diff
@@ -209,6 +218,44 @@ def _auth_with_jwt(self, sub, sub_type):
...
+            `unicode`
+        :return:
+            The access token for the enterprise or app user.
+        :rtype:
+            `unicode`
+        """
+        try:
+            return self._construct_and_send_jwt_auth(sub, sub_type)
+        except BoxOAuthException as ex:
+            error_response = ex.network_response
```
jmoldow: Can this be a helper method?

```python
try:
    return self._construct_and_send_jwt_auth(sub, sub_type)
except BoxOAuthException as ex:
    box_datetime = self._get_date_header(ex.network_response)
    if (box_datetime is not None) and self._exp_claim_rejected_due_to_clock_skew(ex.network_response):
        return self._construct_and_send_jwt_auth(sub, sub_type, box_datetime)
    raise

@staticmethod
def _get_date_header(network_response):
    """Get datetime object for Date header, if it is available.

    Returns None if Date header is absent or can't be parsed.
    """
    if 'Date' not in network_response.headers:
        return None
    box_date_header = network_response.headers['Date']
    try:
        return datetime.strptime(box_date_header, '%a, %d %b %Y %H:%M:%S %Z')
    except ValueError:
        return None

@staticmethod
def _exp_claim_rejected_due_to_clock_skew(network_response):
    status_code = error_response.status_code
    try:
        json_response = error_response.json()
    except ValueError:
        return False
    error_code = json_response.get('error', '')
    error_description = json_response.get('error_description', '')
    return (status_code == 400) and (error_code == 'invalid_grant') and ('exp' in error_description)
```

boxsdk/auth/jwt_auth.py
```diff
@@ -209,6 +218,44 @@ def _auth_with_jwt(self, sub, sub_type):
...
+            try:
+                status_code = error_response.status_code
+                json_response = error_response.json()
+                error_description = json_response['error_description']
+                box_date_header = error_response.headers['date']
+                box_datetime = datetime.strptime(box_date_header, '%a, %d %b %Y %H:%M:%S %Z')
+            except Exception:  # pylint:disable=broad-except
+                pass
+            else:
+                if status_code == 400 and 'exp' in error_description:
```
jmoldow: Can we also test for `json_response['error'] == 'invalid_grant'`? I don't like solely depending on a human-readable string.

PR #253, opened 2017-11-29
Rework automatic retry.

Jeff-Meadows:
Previously, retries were done recursively, which could lead to large
stacks and even stack overflow. This commit comprises a refactoring
of automatic retry.

* BoxResponse is moved to its own module (from box_session)
* A new class, BoxRequest, exists to keep track of information about
  a specific API request while it's being made or retried.
* BoxSession was refactored so that retries happen iteratively:
  - _make_request was renamed _send_request and takes a BoxRequest.
  - _retry_request_if_necessary was renamed _get_retry_request_callable
    and it now returns a callable that can retry a request, rather
    than actually retrying it.
  _prepare_and_send_request now retries failed requests in a loop
* The retry strategy has changed so that 429 responses won't be
  retried indefinitely - we'll give up after 10 failures regardless
  of whether they are 5xx responses or 429s.

### DIFF
--- a/boxsdk/session/box_response.py
+++ b/boxsdk/session/box_response.py
@@ -1,0 +1,58 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+
+class BoxResponse(object):
+    """Represents a response to a Box API request."""
+
+    def __init__(self, network_response):
+        self._network_response = network_response
+
+    def json(self):
+        """Return the parsed JSON response.
+
+        :rtype:
+            `dict` or `list` or `str` or `int` or `float`
+        """
+        return self._network_response.json()
+
+    @property
+    def content(self):
+        """Return the content of the response body.
+
+        :rtype:
+            varies
+        """
+        return self._network_response.content
+
+    @property
+    def ok(self):
+        """Return whether or not the request was successful.
+
+        :rtype:
+            `bool`
+        """
+        # pylint:disable=invalid-name
+        return self._network_response.ok
+
+    @property
+    def status_code(self):
+        """Return the HTTP status code of the response.
+
+        :rtype:
+            `int`
+        """
+        return self._network_response.status_code
+
+    @property
+    def network_response(self):
+        """Return the underlying network response.
+
+        :rtype:
+            :class:`NetworkResponse`
+        """
+        return self._network_response
+
+    def __repr__(self):
+        return '<Box Response[{status_code}]>'.format(status_code=self.status_code)
--- a/boxsdk/session/box_session.py
+++ b/boxsdk/session/box_session.py
@@ -1,64 +1,16 @@
 # coding: utf-8
 
-from __future__ import unicode_literals
+from __future__ import unicode_literals, absolute_import
 
-from boxsdk.config import API, Client
-from boxsdk.exception import BoxAPIException
-from boxsdk.util.multipart_stream import MultipartStream
-from boxsdk.util.shared_link import get_shared_link_header
+from functools import partial
+
+from .box_request import BoxRequest as _BoxRequest
+from .box_response import BoxResponse as _BoxResponse
+from ..config import API, Client
+from ..exception import BoxAPIException
+from ..util.multipart_stream import MultipartStream
+from ..util.shared_link import get_shared_link_header
 from ..util.translator import Translator
-
-
-class BoxResponse(object):
-    """Represents a response to a Box API request."""
-
-    def __init__(self, network_response):
-        self._network_response = network_response
-
-    def json(self):
-        """Return the parsed JSON response.
-
-        :rtype:
-            `dict` or `list` or `str` or `int` or `float`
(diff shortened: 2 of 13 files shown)


boxcla:

boxsdk/session/box_response.py
```diff
@@ -0,0 +1,58 @@
...
+            varies
+        """
+        return self._network_response.content
+
+    @property
+    def ok(self):
+        """Return whether or not the request was successful.
+
+        :rtype:
+            `bool`
```
potrebic: Why are these docstrings on 2 lines? vs box_request.py?

boxsdk/session/box_session.py
```diff
@@ -105,6 +57,16 @@ def __init__(self, oauth, network_layer, default_headers=None, translator=None,
...
             self._default_network_request_kwargs.update(default_network_request_kwargs)
 
+    @property
+    def box_request_constructor(self):
+        """Get the constructor for the container class representing an API request"""
+        return _BoxRequest
+
+    @property
+    def box_response_constructor(self):
+        """Get the constructor for the container class representing an API response"""
```
potrebic: :retype: doctoring?

For both of these properties?

boxsdk/session/box_session.py
```diff
@@ -211,14 +173,38 @@ def _is_json_response(network_response):
...
+    def _get_retry_after_time(self, attempt_number, retry_after_header):
         """
-        Retry a request for certain types of failure.
+        Get the amount of time to wait before retrying the API request.
+
+        If the Retry-After header is supplied, use it; otherwise, use exponential backoff
+        For 202 Accepted (thumbnail or file not ready) and 429 (too many requests), retry later, after a delay
+        specified by the Retry-After header.
+        For 5xx Server Error, retry later, after a delay; use exponential backoff to determine the delay.
+
```
potrebic: The doctoring talks about 202's, 500's, etc. But the code below could care less. The code is purely based on retry_after_header or not. 

It seems the doctoring, talking about status_codes, is really documenting some higher level logic elsewhere in the code?

boxsdk/session/box_session.py
```diff
@@ -227,48 +213,39 @@ def _retry_request_if_necessary(self, network_response, attempt_number, *args, *
...
-        :param method:
-            The HTTP verb used to make the request.
-        :type method:
-            `unicode`
-        :param url:
-            The request URL.
-        :type url:
-            `unicode`
+        :param request:
+            The API request that could be unsuccessful.
```
potrebic: suggest:
> ... that could have been unsuccessful ...

boxsdk/session/box_session.py
```diff
@@ -336,68 +312,57 @@ def _prepare_and_send_request(
             `int`
         """
         files = kwargs.get('files')
-        file_stream_positions = None
+        kwargs['file_stream_positions'] = None
```
potrebic: I know this is existing code... a reference to `files` and `file_stream_positions` seems out of the blue here in 100% generic "make a request" code. Might there be a comment explaining why this code is here?

boxsdk/session/box_session.py
```diff
@@ -211,14 +173,38 @@ def _is_json_response(network_response):
...
+    def _get_retry_request_callable(self, network_response, attempt_number, request):
+        """
+        Get a callable that retries a request for certain types of failure.
+
         For 401 Unauthorized responses, renew the session by refreshing the access token; then retry.
         For 202 Accepted (thumbnail or file not ready) and 429 (too many requests), retry later, after a delay
         specified by the Retry-After header.
         For 5xx Server Error, retry later, after a delay; use exponential backoff to determine the delay.
 
+        Otherwise, return None.
```
potrebic: I might make this comment first. Most calls succeed. So mostly this method will return None.

Also makes me wonder if this should be called `def _get_retry_request_on_retriable_failures`
or `_get_retry_request_callable_on_retriable_failures` or `_get_retry_callable_on_retriable_failures`

boxsdk/session/box_session.py
```diff
@@ -336,68 +312,57 @@ def _prepare_and_send_request(
...
-            expect_json_response=True,
-            attempt_number=0,
-            **kwargs
-    ):
+        network_response = self._send_request(request, **kwargs)
+
+        while True:
+            retry = self._get_retry_request_callable(network_response, attempt_number, request)
+
+            if retry is None or attempt_number >= 10:
```
potrebic: The commit message was a bit confusing:
```
we'll give up after 10 failures regardless
of whether they are 5xx responses or 429s
```
because 202's and 401's are in play. The code here gives on up an y `retriable`failure after 10 attempts

boxsdk/session/box_response.py
Jeff-Meadows: Because this code is just moved from the `box_session` module, and `box_request` is new code.

PR #254, opened 2017-11-29
Fix the build for python 2.6 and python 3.3.

Jeff-Meadows:
The latest version of pytest (3.3.0) dropped support for these
python versions, so we need to install an older version when
we're being run on one of them.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,12 +1,1 @@
--r requirements.txt
-bottle
-jsonpatch
-mock>=2.0.0
-pep8
-pylint
-pytest>=2.8.3
-pytest-cov
-pytest-xdist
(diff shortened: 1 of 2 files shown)


boxcla:

jmoldow:
More and more tools are going to be dropping 2.6 support in the future (e.g. pip). At some point we should consider whether we should also drop 2.6 support.

Rather than having conditional Python code inside setup.py, I feel like it would've been better to either:
- Pin to the same `pytest` version for all Python versions, for consistent behavior
- Use a requirements conditional, e.g.

```
pytest>=2.8.3 ; python_version == '2.7' or python_version > '3.3'
pytest>=2.8.3,<3.3.0 ; python_version == '2.6' or python_version == '3.3'
```

PR #263, opened 2018-03-29
Drop support for py2.6 and py3.3.

Jeff-Meadows:
This commit removes testing for those platforms.
This commit also makes changes so that CI passes again, including
several pylint warning suppressions.
This commit also starts using pycodestyle instead of the
deprecated pep8 package name.


boxcla:

PR #262, opened 2018-03-27
Updated make_request per iss261

jpan-box:
Fixes #261

PR #265, opened 2018-03-29
updated make_request to include info about passing a body

### LINKED ISSUE #261: Improve documentation for sending body in make_requests
Hey team, Boxer here :) submitting PR for this shortly.

The helper function here:
https://github.com/box/box-python-sdk/blob/1.5/README.rst#other-requests

Demonstrates how to specify a method and URL for a custom API call. However, it is not clear how to send the body.

Looking closer at the docs:
https://github.com/box/box-python-sdk/blob/f77cb41afe91f1f1d9b7d6b0044ae5eee358c977/boxsdk/client/client.py#L298

It looks like it takes **kwargs, which is some sort of parameter handling. As a beginner, it's not easy to understand how to send a body for a custom API call.

See PR for suggested doc updates...

jpan-box:
Fixes #261 , but now in master instead of 1.5

Duplicates #262 , which will be closed since there are merging errors with 1.5

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -276,6 +276,19 @@ supported by the SDK. You can still use these endpoints by using the ``make_requ
 The ``Client`` class and Box objects have a ``get_url`` method. Pass it an endpoint
 to get the correct URL for use with that object and endpoint.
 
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @jpan-box has signed the CLA. Thanks for the pull request!

Issue #190, opened 2017-02-01
Logging doesn't occur on certain errors


jmoldow:
In #172, when making the logging of network responses be a bit lazy, I noted:

> In theory, this could make the logs less useful, by adding a delay between
> when the network response was actually received, and when it is logged. Or
> the response may never be logged, if the content is never accessed. In
> practice, this is unlikely to happen, because nearly all SDK methods
> immediately read the content.

It turns out that this assessment wasn't quite correct. Most SDK methods do immediately read the content.. on success. On failure, the session doesn't always read the content. For example:

- For most auth failures, we don't read the error response before attempting a refresh.
- For retries on 429 and 5xx responses, we don't read the error response before attempting a retry.

If the SDK knows that it is throwing away a request, it should read the content first. That way, if the logging network is being used, it can be logged.

Alternatively, we can call `del response`, and then have a `__del__` on `LoggingNetworkResponse` which does the content reading. That way, the `BoxSession` code doesn't need to be too closely tied to the logging network (though still a bit tied, because that's the only reason for having `del response`).

PR #277, opened 2018-06-02
Refactor logging

Jeff-Meadows:
This commit refactors logging from a special `LoggingNetwork`
subclass into the `DefaultNetwork` base class.

In addition, this commit fixes `setup_logging` so that calling
it multiple times does not add additional handlers.

The special `LoggingClient` subclass now does nothing more than
call `setup_logging`, since it doesn't need to use a special
network subclass.


boxcla:

boxsdk/session/box_session.py
```diff
@@ -376,6 +378,8 @@ def _prepare_and_send_request(
                 break
 
             attempt_number += 1
+            network_response.log(can_safely_log_content=True)
```
jmoldow: Looks like this partially addresses issue #190 .

As another idea, what if we put this into the constructor for `DefaultNetworkResponse`.

```python
def __init__(self, request_response, access_token_used):
    ....
    if not self.ok:
        self.log(can_safely_log_content=True)
```

(or we could put equivalent code into `DefaultNetwork.request()`)

The idea being that, if the request fails, then in theory the content is never going to be streamed (we could even enforce that if we wanted).  And so in those cases, it is safe to log immediately, without waiting to see how the caller is going to access the response data.

In theory, this would completely address #190.

boxsdk/util/log.py
```diff
@@ -1,12 +1,42 @@
...
+        if not self._has_setup:
+            self._has_setup = True
+            return self._setup_logging(stream_or_file, debug, name)
+        return None
+
+    @staticmethod
+    def _setup_logging(stream_or_file=None, debug=False, name=None):
+        logger = logging.getLogger(name)
+        if isinstance(stream_or_file, string_types):
+            logger.addHandler(logging.FileHandler(stream_or_file, mode='w'))
```
jmoldow: Should it be append mode?

boxsdk/util/log.py
```diff
@@ -1,12 +1,42 @@
...
 from six import string_types
 
 
+class Logging(object):
+    _has_setup = False
+
+    def setup_logging(self, stream_or_file=None, debug=False, name=None):
+        if not self._has_setup:
+            self._has_setup = True
+            return self._setup_logging(stream_or_file, debug, name)
```
jmoldow: Why do we need to return a logger? Can't this be a method that is used 100% for side-effects, and have clients always use the standard `logger.getLogger` to get the logger?

The fact that it can also return `None` means that you can't really rely on the return value anyway.

It looks like all the code in this PR is already doing that, we never access the return value except in tests.

It looks like removing the return value would be a breaking change, but I think that's fine. It's easy enough to deal with the change, and I doubt many people were using this function anyway.

PR #279, opened 2018-06-07
Improve exceptions.

### LINKED ISSUE #117: unicode in BoxException
Currently, `BoxException` is defined like this:

``` python
class BoxException(Exception):
    """
    Base class exception for all errors raised from the SDK.
    """
    def __str__(self):
        return self.__unicode__().encode('utf-8') if PY2 else self.__unicode__()
```

This causes infinite recursion on Python2, because the base class `Exception` (as a subclass of `object`) has a `__unicode__` method that references `__str__`. It is, however, fine in the subclass `BoxAPIException` and `BoxOAuthException`, because they define their own `__unicode__` methods.

And this code will break on Python 3 for the bare `BoxException` class and its subclass `BoxNetworkException`, because by default there is no `__unicode__` method on Python 3 (since it has no meaning in that language).

I've started on a refactor to fix this.

### LINKED ISSUE #145: Make the error code accessible in BoxOAuthException
In the `BoxOAuthException`, both the error code and error description are both stored in a stringified dictionary. This makes it so that in order check if the oauth exception is an `invalid_request` one would have to do `'invalid_request' in oauth_exception._message` instead of something like `'invalid_request' == oauth_exception.code`.

The `BoxAPIException` separates and exposes the error code and description, it seems like it could be useful to do the same for `BoxOAuthException`.

### LINKED ISSUE #164: Censor secrets from logs and exception tracebacks
Information such as:
- client_id
- client_secret
- access_token
- refresh_token
- (less important) names, emails, descriptions, and shared links

should be censored or truncated when logged or put into exception tracebacks. That way, users who copy/paste this information don't risk exposing this sensitive information.

A social solution to this, would be to create a template for Github issues, instructing users to self-censor before copy/pasting anything into Github. https://help.github.com/articles/creating-an-issue-template-for-your-repository/

### LINKED ISSUE #270: Bring exception hierarchy in line with other Box official SDKs

### LINKED ISSUE #276: `boxsdk.exception.BoxAPIException` Documentation Error
In [boxsdk/exception.py](../blob/master/boxsdk/exception.py) (line 115 in `master`, but same problem in `1.5`), the documentation of `boxsdk.exception.BoxAPIException` says the return type of the `.code` property is `int`, but I believe it is actually `unicode`. Presumably a copy/paste from the immediately-above `.status` property, whose return type really is `int`.

Additionally, the type of the `code` parameter to the `__init__` of the same class is documented as `unicode` earlier in the file.

I have found the [attrs](http://www.attrs.org/en/stable/) library useful.

Jeff-Meadows:
Fixes #276
Fixes #270
Fixes #164
Fixes #145
Fixes #117

This commit refactors the Box exception classes to have more useful
string representations.

It also sanitizes sensitive information out of the exceptions and
out of our logging.

Finally, it fixes up some of the documentation for the exception
classes.

### DIFF
--- a/boxsdk/__init__.py
+++ b/boxsdk/__init__.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals, absolute_import
 
 from .auth import JWTAuth, OAuth2
 from .client import *  # pylint:disable=wildcard-import,redefined-builtin
+from .exception import *  # pylint:disable=wildcard-import
 from .object import *  # pylint:disable=wildcard-import,redefined-builtin
 from .util.log import setup_logging
 from .version import __version__
--- a/boxsdk/auth/jwt_auth.py
+++ b/boxsdk/auth/jwt_auth.py
@@ -16,7 +16,7 @@ from six import binary_type, string_types, raise_from, text_type
 from ..exception import BoxOAuthException
 from .oauth2 import OAuth2
 from ..object.user import User
-from ..util.compat import NoneType, total_seconds
+from ..util.compat import NoneType
 
 
 class JWTAuth(OAuth2):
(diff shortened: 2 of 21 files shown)


boxcla:

PR #280, opened 2018-06-16
Add Comments functionality

mattwiller:


boxcla:

test/unit/object/test_comment.py
```diff
@@ -0,0 +1,95 @@
...
+        'type': 'comment',
+        'id': '12345',
+        message_type: message
+    }
+    reply_comment = test_comment.reply(message)
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_data))
+    assert reply_comment.object_id == '12345'
+
+
+@pytest.mark.parametrize(
```
Jeff-Meadows: This is copy pasta from the previous test - can we either put this into a fixture or just pull out the parametrize call, like

    comment_parameters = @pytest.mark.parametrize(...)

    @comment_parameters
    def test_edit(...)

test/unit/object/test_file.py
```diff
@@ -278,3 +278,61 @@ def test_get_shared_link_download_url(
...
+        }
+    }
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'comment',
+        'id': '12345',
+        message_type: message
+    }
+    comment = test_file.add_comment(message)
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_data))
+    assert comment.object_id == '12345'
```
Jeff-Meadows: pull out `'12345'` into a variable

test/unit/object/test_comment.py
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'comment',
+        'id': '12345',
+        message_type: message
+    }
+    reply_comment = test_comment.reply(message)
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_data))
+    assert reply_comment.object_id == '12345'
```
jmoldow: Maybe assert that you got a `Comment` back?

test/unit/object/test_comment.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_edit(test_comment, mock_box_session, comment_params):
+    expected_url = mock_box_session.get_url('comments', test_comment.object_id)
+    (message_type, message) = comment_params
+    expected_data = {
+        message_type: message,
+    }
+    mock_box_session.post.return_value.json.return_value = {
```
jmoldow: Shouldn't this be `put`, not `post`?

Apparently it doesn't matter since it didn't cause a test failure. But that's probably because there aren't any assertions on the response type/data.

test/unit/object/test_file.py
```diff
@@ -278,3 +278,47 @@ def test_get_shared_link_download_url(
...
+                'message': 'Foo'
+            },
+            {
+                'type': 'comment',
+                'id': '2',
+                'message': 'Bar'
+            }
+        ]
+    }
+    test_file.get_comments().next()
```
jmoldow: We can do more than this. Why not assert that we get two comments back with the expected comments?

test/unit/object/test_file.py
```diff
@@ -278,3 +278,47 @@ def test_get_shared_link_download_url(
...
+        }
+    }
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'comment',
+        'id': comment_id,
+        message_type: message
+    }
+    comment = test_file.add_comment(message)
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_data))
+    assert comment.object_id == comment_id
```
jmoldow: Can we assert that this is a `Comment`?

boxsdk/object/comment.py
```diff
@@ -12,6 +12,14 @@ class Comment(BaseObject):
     """An object that represents a comment on an item"""
     _item_type = 'comment'
 
+    @staticmethod
+    def construct_params_from_message(message):
+        message_type = 'tagged_message' if '@[' in message else 'message'
+        data = {
```
jmoldow: In general, if you're going to do

```python
foo = ...
return foo
```

you can just replace it with 

```python
return ...
```

boxsdk/object/comment.py
```diff
@@ -23,13 +31,10 @@ def reply(self, message):
...
         """
         url = self.get_type_url()
-        message_type = 'tagged_message' if '@[' in message else 'message'
-        data = {
-            message_type: message,
-            'item': {
-                'type': 'comment',
-                'id': self.object_id
-            }
+        data = Comment.construct_params_from_message(message)
```
jmoldow: This should be `self.construct_params_from_message`.

staticmethods and classmethods can be called with `self`. Python will do the correct thing.

Hard-coding the current class should almost always be avoided, since it prevents overriding the behavior in a subclass.

Another thing that would work (but isn't necessary) is `self.__class__.construct_params_from_message`, but it shouldn't be any different from `self.construct_params_from_message`.

boxsdk/object/comment.py
```diff
@@ -49,8 +54,5 @@ def edit(self, message):
         :type message:
             `unicode`
         """
-        message_type = 'tagged_message' if '@[' in message else 'message'
-        data = {
-            message_type: message
-        }
+        data = Comment.construct_params_from_message(message)
```
jmoldow: See above.

boxsdk/object/file.py
```diff
@@ -327,13 +328,10 @@ def add_comment(self, message):
...
         """
         url = self._session.get_url('comments')
-        message_type = 'tagged_message' if '@[' in message else 'message'
-        data = {
-            message_type: message,
-            'item': {
-                'type': 'file',
-                'id': self.object_id
-            }
+        data = Comment.construct_params_from_message(message)
```
jmoldow: I think it might be better to do

```python
CommentClass = self._session.translator.translate('comment')
CommentClass.construct_params_from_message(message)
```

But I don't have strong opinions. I'd be fine if you left it like this for now. It's not that likely that someone will need to override the behavior.

jmoldow:
@mattwiller I left some more comments, but I think I did it wrong. The Github UI makes them look like they're outdated, but they are in fact new comments.

PR #281, opened 2018-06-19
Add Collections functionality

mattwiller:


boxcla:

boxsdk/client/client.py
```diff
@@ -143,6 +144,44 @@ def collaboration(self, collab_id):
...
+        :param offset:
+            The user index at which to start the response.
+        :type offset:
+            `int`
+        """
+        return LimitOffsetBasedObjectCollection(
+            self.session,
+            self._session.get_url('collections'),
+            limit=limit,
+            fields=None,
```
jmoldow: Add `fields` to method params.

boxsdk/object/item.py
```diff
@@ -356,3 +356,70 @@ def metadata(self, scope='global', template='properties'):
...
+        }
+        box_response = self._session.get(url, params=params)
+        response = box_response.json()
+
+        collections = response['collections']
+        collections.append({'id': collection.object_id})
+        data = {
+            'collections': collections
+        }
+        box_response = self._session.put(url, data=json.dumps(data))
```
jmoldow: Everything here and below is done automatically for you with `self.update_info()` (inherited from `BaseObject`).

boxsdk/object/item.py
```diff
@@ -356,3 +356,70 @@ def metadata(self, scope='global', template='properties'):
...
+            'fields': 'collections'
+        }
+        box_response = self._session.get(url, params=params)
+        response = box_response.json()
+
+        collections = [c for c in response['collections'] if c['id'] != collection.object_id]
+        data = {
+            'collections': collections
+        }
+        box_response = self._session.put(url, data=json.dumps(data))
```
jmoldow: Use `self.update_info()`.

boxsdk/object/item.py
```diff
@@ -356,3 +356,70 @@ def metadata(self, scope='global', template='properties'):
...
+        :type collection:
+            :class:`Collection`
+        :return:
+            This item.
+        :rtype:
+            :class:`Item`
+        """
+        url = self.get_url()
+        params = {
+            'fields': 'collections'
```
jmoldow: Perhaps we should also get the `etag` and `sequence_id` and pass the correct `If-Match` header, that way we don't have race conditions with the PUT?

Either that, or give a warning in the docstring that this method isn't safe from race conditions.

Same for `remove_from_collection`.

test/unit/object/conftest.py
```diff
@@ -98,6 +104,12 @@ def mock_user(mock_box_session, mock_user_id):
     return user
 
 
+@pytest.fixture()
+def mock_collection(mock_box_session, mock_collection_id):
+    collection = Collection(mock_box_session, mock_collection_id)
+    return collection
```
jmoldow: This can just be a single line.

test/unit/object/test_item.py
```diff
@@ -141,3 +141,43 @@ def test_get(test_item_and_response, mock_box_session, fields, mock_object_id, e
...
+def test_add_to_collection(test_item_and_response, mock_box_session, mock_collection):
+    # pylint:disable=redefined-outer-name, protected-access
+    test_item, mock_item_response = test_item_and_response
+    expected_url = test_item.get_url()
+    expected_params = {'fields': 'collections'}
+    expected_data = {
+        'collections': [{'id': mock_collection.object_id}]
+    }
+    mock_response = {
+        'collections': []
```
jmoldow: Can you add tests where:
- The collection is already in the list.
- A different collection is already in the list.

test/unit/object/test_item.py
```diff
@@ -141,3 +141,43 @@ def test_get(test_item_and_response, mock_box_session, fields, mock_object_id, e
...
+def test_remove_from_collection(test_item_and_response, mock_box_session, mock_collection):
+    # pylint:disable=redefined-outer-name, protected-access
+    test_item, mock_item_response = test_item_and_response
+    expected_url = test_item.get_url()
+    expected_params = {'fields': 'collections'}
+    expected_data = {
+        'collections': []
+    }
+    mock_response = {
+        'collections': [{'id': mock_collection.object_id}]
```
jmoldow: Can you add tests where:
- The collections list is empty.
- The collections list is non-empty, but doesn't contain the given collection.
- The collections list contains the given collection and another collection.

boxsdk/object/item.py
```diff
@@ -356,3 +356,58 @@ def metadata(self, scope='global', template='properties'):
...
+        }
+        box_response = self._session.get(url, params=params)
+        response = box_response.json()
+
+        collections = response['collections']
```
jmoldow: Didn't notice this before, but you can also replace this with

```python
collections = self.get(fields=['collections']).collections
```

boxsdk/object/item.py
```diff
@@ -356,3 +356,58 @@ def metadata(self, scope='global', template='properties'):
...
+            This item.
+        :rtype:
+            :class:`Item`
+        """
+        url = self.get_url()
+        params = {
+            'fields': 'collections'
+        }
+        box_response = self._session.get(url, params=params)
+        response = box_response.json()
```
jmoldow: Ditto here, you can use `self.get`

test/unit/object/test_item.py
```diff
@@ -12,6 +12,30 @@ def test_item_and_response(test_file, test_folder, mock_file_response, mock_fold
...
+    if request.param == 'empty':
+        return [], [{'id': mock_collection_id}]
+    elif request.param == 'same':
+        return [{'id': mock_collection_id}], [{'id': mock_collection_id}, {'id': mock_collection_id}]
+
+    return [{'id': other_collection_id}], [{'id': other_collection_id}, {'id': mock_collection_id}]
+
+
+@pytest.fixture(params=('empty', 'only_removed', 'only_other', 'other_and_removed'))
+def test_collections_for_removal(mock_collection_id, request):
```
jmoldow: Just because it's not immediately obvious, can you add a really short docstring just giving a better idea of what the two tuple elements represent?

test/unit/object/test_item.py
```diff
@@ -12,6 +12,30 @@ def test_item_and_response(test_file, test_folder, mock_file_response, mock_fold
     return test_folder, mock_folder_response
 
 
+@pytest.fixture(params=('empty', 'same', 'other'))
+def test_collections_for_addition(mock_collection_id, request):
```
jmoldow: Just because it's not immediately obvious, can you add a really short docstring just giving a better idea of what the two tuple elements represent?

Same for the next fixture.

test/unit/object/test_item.py
```diff
@@ -12,6 +12,30 @@ def test_item_and_response(test_file, test_folder, mock_file_response, mock_fold
...
+    other_collection_id = mock_collection_id + '2'
+    if request.param == 'empty':
+        return [], [{'id': mock_collection_id}]
+    elif request.param == 'same':
+        return [{'id': mock_collection_id}], [{'id': mock_collection_id}, {'id': mock_collection_id}]
```
jmoldow: Don't we want the expected collections to just be:

```python
[{'id': mock_collection_id}]
```

test/unit/object/test_item.py
```diff
@@ -12,6 +12,30 @@ def test_item_and_response(test_file, test_folder, mock_file_response, mock_fold
...
+
+    return [{'id': other_collection_id}], [{'id': other_collection_id}, {'id': mock_collection_id}]
```
jmoldow: My personal preference would be to have something like:

```python
elif request.param == 'other':
    return [....]

throw NotImplementedError("Forgot to implement {}".format(request.param))
```

Same for the next fixture too.

boxsdk/object/item.py
```diff
@@ -356,3 +356,58 @@ def metadata(self, scope='global', template='properties'):
             :class:`Metadata`
         """
         return Metadata(self._session, self, scope, template)
+
+    @api_call
+    def add_to_collection(self, collection):
+        """
+        Add the item to a collection.
```
jmoldow: Can you warn about the race condition?

boxsdk/object/item.py
```diff
@@ -356,3 +356,58 @@ def metadata(self, scope='global', template='properties'):
...
+        data = {
+            'collections': collections
+        }
+        return self.update_info(data)
+
+    @api_call
+    def remove_from_collection(self, collection):
+        """
+        Remove the item from a collection.
+
```
jmoldow: Can you warn about the race condition?

boxsdk/object/item.py
```diff
@@ -356,3 +356,45 @@ def metadata(self, scope='global', template='properties'):
...
+        :param collection:
+            The collection to add the item to.
+        :type collection:
+            :class:`Collection`
+        :return:
+            This item.
+        :rtype:
+            :class:`Item`
+        """
+        collections = self.get(fields=['collections']).collections  # pylint:disable=no-member
```
jmoldow: I don't care either way, but another approach would be to use `['collections']` instead of `.collections` so that we don't have to worry about silencing the pylint error. Ditto below.

PR #283, opened 2018-06-20
Fix API object equality test

### LINKED ISSUE #163: BaseObject.__eq__ is incorrect
`BaseObject.__eq__` only compares object ID, not type. So `client.file(id) == client.folder(id)` evaluates to `True`.

It should also return `NotImplemented` if the other object is not a `BaseObject`, but right now it will raise an exception (e.g. #151 ).

This is technically a backwards-incompatible change. We'll include this in the major version bump v2.0.0.

We also need to define `__hash__`.

I've started work on this already.

mattwiller:
Fixes #163

### DIFF
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -155,8 +155,24 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
         return box_response.ok
 
     def __eq__(self, other):
-        """Equality as determined by object id"""
-        return self._object_id == other.object_id
+        """Equality as determined by object id and type"""
+        if (other is None):
+            return False
+        if not isinstance(other, BaseObject):
+            return False
+
+        # Two objects are considered the same if they have the same address in the API
+        return self.get_url() == other.get_url()
+
+    def __ne__(self, other):
+        """Equality as determined by object id and type"""
+
+        # Inequality is the inverse of equality
+        return not self == other
+
+    def __hash__(self):
+        """Custom object hash"""
+        return hash((self._object_id, self._item_type))
 
     def _paging_wrapper(self, url, starting_index, limit, factory=None):
         """
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -156,11 +156,12 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
 
     def __eq__(self, other):
         """Equality as determined by object id and type"""
-        if not isinstance(other, BaseObject):
-            return False
+        if isinstance(other, BaseObject):
+            # Two objects are considered the same if they have the same address in the API
+            return self.get_url() == other.get_url()
 
-        # Two objects are considered the same if they have the same address in the API
-        return self.get_url() == other.get_url()
+        return NotImplemented
+
 
     def __ne__(self, other):
         """Equality as determined by object id and type"""
--- a/boxsdk/object/base_object.py
+++ b/boxsdk/object/base_object.py
@@ -162,7 +162,6 @@ class BaseObject(BaseEndpoint, BaseAPIJSONObject):
 
         return NotImplemented
 
-
     def __ne__(self, other):
         """Equality as determined by object id and type"""
         return not self == other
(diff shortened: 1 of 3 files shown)


boxcla:

boxsdk/object/base_object.py
```diff
@@ -155,8 +155,24 @@ def delete(self, params=None, headers=None):
...
-        """Equality as determined by object id"""
-        return self._object_id == other.object_id
+        """Equality as determined by object id and type"""
+        if (other is None):
+            return False
+        if not isinstance(other, BaseObject):
+            return False
+
+        # Two objects are considered the same if they have the same address in the API
+        return self.get_url() == other.get_url()
```
jmoldow: Cool approach. I hadn't thought of doing it this way, but this makes a lot of sense.

One thing to note about this approach (which is probably, but not necessarily, an upside) is that two objects will compare as unequal if they are the same except for the version of the API that they are pointing at.

boxsdk/object/base_object.py
```diff
@@ -155,8 +155,24 @@ def delete(self, params=None, headers=None):
         return box_response.ok
 
     def __eq__(self, other):
-        """Equality as determined by object id"""
-        return self._object_id == other.object_id
+        """Equality as determined by object id and type"""
+        if other is None:
```
jmoldow: None is valid to pass to `isinstance()`, so this check is redundant with the next check and is unnecessary.

boxsdk/object/base_object.py
```diff
@@ -155,8 +155,24 @@ def delete(self, params=None, headers=None):
...
+            return False
+        if not isinstance(other, BaseObject):
+            return False
+
+        # Two objects are considered the same if they have the same address in the API
+        return self.get_url() == other.get_url()
+
+    def __ne__(self, other):
+        """Equality as determined by object id and type"""
+
```
jmoldow: Unnecessary blank line.

boxsdk/object/base_object.py
```diff
@@ -155,8 +155,24 @@ def delete(self, params=None, headers=None):
...
+        # Inequality is the inverse of equality
```
jmoldow: This is always the case, doesn't need a comment.

boxsdk/object/base_object.py
```diff
@@ -155,8 +155,24 @@ def delete(self, params=None, headers=None):
...
+        return not self == other
+
+    def __hash__(self):
+        """Custom object hash"""
```
jmoldow: Docstring is unnecessary here. `__hash__` is always a "custom object hash"

test/unit/object/test_base_object.py
```diff
@@ -21,6 +23,19 @@ def test_object_and_response(
     return test_objects_and_responses[request.param]
 
 
+@pytest.fixture(params=('same_file', 'same_folder', 'same_user', 'against_none', 'different_ids', 'different_types'))
+def objects_for_comparison(test_file, test_folder, mock_user, request, mock_box_session):
+    cases = {
+        'same_file': (test_file, test_file, True),
```
jmoldow: Because of the way you've defined these first three fixtures, your tests are trivially correct. It would be difficult (though obviously not impossible) to accidentally write an incorrect equality such that `foo == foo` isn't `True`.

The true test of your equality would be if `fooFactory() == fooFactory()` is `True`.

So here, I think what we really want is
```python
'same_file': (test_file, File(test_file.object_id), True),
...
```

test/unit/object/test_base_object.py
```diff
@@ -21,6 +24,21 @@ def test_object_and_response(
...
+def objects_for_comparison(test_file, test_folder, mock_user, request, mock_box_session):
+    cases = {
+        'identical_object': (test_file, test_file, True),
+        'same_file': (test_file, File(mock_box_session, test_file.object_id), True),
+        'same_folder': (test_folder, Folder(mock_box_session, test_folder.object_id), True),
+        'same_user': (mock_user, User(mock_box_session, mock_user.object_id), True),
+        'against_none': (test_file, None, False),
+        'different_ids': (File(mock_box_session, '1'), File(mock_box_session, '2'), False),
+        'different_types': (File(mock_box_session, '1'), Folder(mock_box_session, '1'), False),
+        'not_api_object': (test_file, {'id': test_file.object_id, 'type': 'file'}, False)
```
jmoldow: We prefer to use trailing commas (which is supported in Python in almost all cases), because then you can add new elements without affecting the diff of the previously-last line.

PR #285, opened 2018-06-20
Skip CI on feature branch push

mattwiller:
In order to make CI run more quickly for PRs, we can skip running tests on push to feature branches (since those will be tested by the PR CI job already).  Pushes on "stable" branches will still be run.


boxcla:

PR #288, opened 2018-06-25
Use paging iterators for all collection-returning methods

### LINKED ISSUE #274: Finish making breaking changes to paging APIs
We should be using the new paging wrappers.

Also do one final audit of the new paging wrappers to make sure that the interface is what we want.

mattwiller:
Fixes #274


boxcla:

PR #290, opened 2018-06-26
Add type check in logging sanitizer

mattwiller:
Log sanitization should not attempt to sanitize non-string values,
since they are not guaranteed to be subscriptable.


jmoldow:
We can add a unit test for this.

PR #295, opened 2018-06-29
Remove Python 2.6 and 3.3 from test matrix

mattwiller:


jmoldow:
Please also update README and CHANGELOG.

While we're at it, we should consider adding 3.7 to the test matrix. It was released yesterday.

mattwiller:
@jmoldow I updated the README and Changelog, but it doesn't look like 3.7 has been added to TravisCI yet

PR #303, opened 2018-07-19
Legal holds

carycheng:


boxcla:

boxsdk/object/legal_hold_policy.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_url(self, *args):
+        return self._session.get_url('legal_hold_policies', self._object_id, *args)
+
+    def assign(self, item):
+        url = self._session.get_url('legal_hold_policy_assignments')
+        body = {
+            'policy_id': self.object_id,
+            'assign_to': {
+                'type': item.type,
```
mattwiller: Does this need to be `item.object_type`?

test/unit/object/test_legal_hold_policy.py
```diff
@@ -0,0 +1,112 @@
...
+    return 101
+
+
+@pytest.fixture(scope='module')
+def policy_id_2():
+    return 202
+
+
+@pytest.fixture(scope='module')
+def policies_response(policy_id_1, policy_id_2):
```
mattwiller: These fixtures should only be defined in one place

boxsdk/object/legal_hold_policy.py
```diff
@@ -0,0 +1,80 @@
...
+from ..config import API
+from ..pagination.marker_based_object_collection import MarkerBasedObjectCollection
+
+class LegalHoldPolicy(BaseObject):
+    """Represents a Box legal_hold_policy"""
+
+    _item_type = 'legal_hold_policy'
+
+    def get_url(self, *args):
+        return self._session.get_url('legal_hold_policies', self._object_id, *args)
```
jmoldow: Ha, is this the first object we have that isn't pluralized by just adding an "s" at the end?

boxsdk/object/legal_hold_policy.py
```diff
@@ -0,0 +1,80 @@
...
+        :param fields:
+            List of fields to request
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            An iterator of the entries in the legal hold policy assignment
+        """
+        additional_params = {
+            'policy_id': self.object_id,
+            'assign_to_type': assign_to_type,
```
jmoldow: Should these be omitted if they are `None`?

boxsdk/object/legal_hold_policy.py
```diff
@@ -0,0 +1,80 @@
...
+            'assign_to_id': assign_to_id,
```
jmoldow: Should we do additional validation to make sure that if `assign_to_id` is given, that `assign_to_type` is also given? Just passing in an ID by itself is meaningless.

Is `assign_to_type` allowed to be given by itself, or does it always have to go with an ID.

If they always have to be given together, then we can have clients instead pass a `BaseObject` instead of separate type and ID.

boxsdk/pagination/page.py
```diff
@@ -38,7 +39,7 @@ def _translator(self):
 
         :rtype:   :class:`Translator`
         """
-        return self._session.translator
+        return Translator()
```
jmoldow: Please revert this change.

Why did you make this change?

PR #305, opened 2018-07-20
Refactor auth to use the session.

Jeff-Meadows:
Previously, auth classes used the network layer directly, while all
other API calls used the session as a network abstraction. This
was not only confusing, but it robbed auth calls of all the session's
features, like automatic retry, user agent, custom headers, etc.

The reason for this was the SDK's architecture. Since the session
handles auth refresh, it needed an auth instance. This commit
refactors the session into unauthorized and authorized parts. Auth
calls now use the unauthorized session, and other API calls use
the authorized session.


boxcla:

PR #306, opened 2018-07-23
Fully translate nested objects in API responses

mattwiller:


boxcla:

jmoldow:
Is this a WIP, or final code for review?

mattwiller:
@jmoldow This is more like a first draft — this works in my local testing, so it would be great if you could look it over and evaluate if you think this is the right approach!

boxsdk/util/translator.py
```diff
@@ -128,5 +131,40 @@ def translate(self, type_name):
...
+                    'object_id': object_id,
+                }
+                params = inspect.signature(object_class.__init__).parameters
+                param_values = {p:param_values[p] for p in params if self._is_constructor_param(params[p])}
+                return self.get(object_type)(**param_values)
+
+        return response_object
+
+    def _is_constructor_param(self, param):
+        if param.name is 'self':
```
Jeff-Meadows: compare with `==` not `is`

boxsdk/util/translator.py
```diff
@@ -116,17 +134,45 @@ def get(self, key, default=None):
...
+            # The $type value in metadata instances isn't directly usable, so we avoid it altogether
+            # NOTE: Currently, we represent metadata as just a `dict`, so there's no need to translate it anyway
+            if 'type' in response_object and '$type' not in response_object:
+                object_class = self.get(response_object.get('type', ''))
+                param_values = {
+                    'session': session,
+                    'response_object': response_object,
+                    'object_id': _get_object_id(response_object),
+                }
+                # NOTE: getargspec() is deprecated, and should be replaced by inspect.signature() when 2.7 support drops
```
Jeff-Meadows: I hope 2.7 support doesn't outlast support for `getargspec`, but it might. It could be safer or more correct to use `inspect.signature`, falling back to `getargspec` when unavailable.

I think the `mock` library does something similar.

mattwiller:
@Jeff-Meadows It looks like `mock` uses another library — `funcsigs` — when `inspect.signature` is unavailable; I think I can just use `getargspec()` instead though.  I'll look into adding the fallback logic.

Also, your previous comment about this method being more expensive is correct — it takes about 45x longer to fully translate than to shallow translate in the tests I ran (against a collection of 1000 files in a folder, as well as a single file).  I'm not sure how dire that is in context, but the difference I observed (a few hundred microseconds) for each one of a few thousand items in an iterator could certainly add up to non-trivial amounts of time.  I'll also explore how easy it would be to lazily translate nested objects; the case I'm currently worried about is when an `APIJSONObject` like `Event` needs to translate a nested `BaseObject` like `File`.  We may need to have `APIJSONObject` hold onto a `Session` instance even though it itself doesn't require it, just in case a nested object does require it at translation time.

Jeff-Meadows:
@mattwiller one more comment - we should review carefully and perhaps give some suggestions to SDK users about when to use `translator.get` vs `translator.translate`. It seems that we're going to end up using `get` for some API calls that could benefit from calling `translate`.

PR #310, opened 2018-07-27
Tasks

carycheng:


boxcla:

boxsdk/object/file.py
```diff
@@ -289,3 +293,89 @@ def get_shared_link_download_url(
...
+            },
+            'action': 'review'
+        }
+        if message is not None:
+            task_attributes['message'] = message
+        if due_at is not None:
+            task_attributes['due_at'] = due_at
+        box_response = self._session.post(url, data=json.dumps(task_attributes))
+        response = box_response.json()
+        return Task(self._session, response['id'], response)
```
jmoldow: Use the translator.

boxsdk/object/file.py
```diff
@@ -289,3 +293,89 @@ def get_shared_link_download_url(
...
+
+
+    def tasks(self, fields=None):
```
jmoldow: Add `limit` and `offset`, don't hard-code them.

boxsdk/object/task.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def assign(self, assign_to_id=None, assign_to_login=None):
+        url = self._session.get_url('task_assignments')
+        body = {
+            'task': {
+                'type': 'task',
+                'id': self.object_id
+            },
+            'assign_to': {
+                'id': assign_to_id,
```
jmoldow: I assume that these are mutually exclusive, and should not both be specified at once.

I would suggest doing this:

Have just an `assign_to` parameter. It can be a `BaseObject` or a `unicode` string. If it is a `unicode` string, then set `assign_to.login`. Otherwise, set `assign_to.id` to `assign_to.object_id`.

(If the API allows you to pass `assign_to.type`, then for good measure, you should pass `assign_to.object_type` as `assign_to.type`, or else the `"user"` constant if you're passing a login.)

boxsdk/object/task.py
```diff
@@ -0,0 +1,58 @@
...
+                'login': assign_to_login
+            }
+        }
+        response = self._session.post(url, data=json.dumps(body)).json()
```
jmoldow: `self.update_object()`.

boxsdk/object/task.py
```diff
@@ -0,0 +1,58 @@
...
+        return Translator().translate(response['type'])(
+            self._session,
+            response['id'],
+            response,
+        )
+
+    def assignments(self, fields=None):
```
jmoldow: Pass `limit` and `offset`, don't hard-code.

boxsdk/object/task_assignment.py
```diff
@@ -0,0 +1,14 @@
...
+from __future__ import unicode_literals
+
+import json
+
+from .base_object import BaseObject
+from boxsdk.config import API
+
+
+class TaskAssignment(BaseObject):
+    """Represents a Box task."""
```
jmoldow: "task assignment"

boxsdk/object/task_assignment.py
```diff
@@ -0,0 +1,14 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
+
+import json
+
+from .base_object import BaseObject
+from boxsdk.config import API
```
jmoldow: Unneeded import.

boxsdk/object/task_assignment.py
```diff
@@ -0,0 +1,14 @@
+# coding: utf-8
+
+from __future__ import unicode_literals
```
jmoldow: Also add `absolute_import` for all `__future__` imports.

boxsdk/pagination/page.py
```diff
@@ -38,7 +39,7 @@ def _translator(self):
 
         :rtype:   :class:`Translator`
         """
-        return self._session.translator
+        return Translator()
```
jmoldow: Revert

PR #315, opened 2018-07-30
User

carycheng:


boxcla:

boxsdk/object/user.py
```diff
@@ -2,10 +2,86 @@
...
+        :param notify:
+            Whether the destination user should receive email notification of the transfer.
+        :type notify:
+            `boolean`
+        :param fields:
+            List of fields to request.
+        :type fields:
+            `Iterable` of `unicode`
+        """
+        url = '{0}/users/{1}/folders/{2}'.format(API.BASE_API_URL, user_id, '0')
```
jmoldow: No need to format the `'0'`. Just put it directly into the original string.

boxsdk/object/user.py
```diff
@@ -2,10 +2,86 @@
...
+        return MarkerBasedObjectCollection(
+            session=self._session,
+            url='{0}/users/{1}/email_aliases'.format(API.BASE_API_URL, self.object_id),
+            marker=None,
+            fields=fields,
+            return_full_pages=False
+        )
+
+
+    def move_owned_items(self, user_id, notify=None, fields=None):
```
jmoldow: You should be more specific about what the user is. Perhaps call it `target_user` or something.

Also, pass a `User` object, not an ID.

PR #320, opened 2018-08-10
Altered client and box session objects to accept raise_exception parameter

jdeepee:
(True/False) where if False it will not raise exception but still return response


boxcla:
Hi @jdeepee, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

PR #324, opened 2018-08-20
Webhooks v2

carycheng:


boxcla:

boxsdk/client/client.py
```diff
@@ -511,6 +511,88 @@ def get_groups(self, name=None, limit=None, offset=None, fields=None):
...
+        Create a webhook on the given file.
+
+        :param target:
+            Either a :class:`File` or :class:`Folder` to assign a webhook to.
+        :type target:
+            :class:`File` or :class`Folder`
+        :param triggers:
+            Event types that trigger notifications for the target.
+        :type triggers:
+            `list of unicode`
```
mattwiller: Should be formatted like this: \`list\` of \`unicode\`

test/unit/client/test_client.py
```diff
@@ -526,6 +578,103 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+    assert isinstance(new_webhook, Webhook)
+    assert new_webhook.id == test_webhook.object_id
+    assert new_webhook.type == test_webhook.object_type
+    assert new_webhook.target['type'] == test_folder.object_type
+    assert new_webhook.target['id'] == test_folder.object_id
+    assert new_webhook.triggers == ['FILE.UPLOADED', 'FOLDER.CREATED']
+    assert new_webhook.address == 'https://test.com'
+
+
+def test_create_webhook_on_file_returns_the_correct_policy_object(
```
mattwiller: Rather than duplicating this test, it might be better to create a fixture that enumerates both the file and folder fixtures — see the `test_item_and_response` fixture in test_item.py

PR #325, opened 2018-08-20
Watermaking v2

carycheng:


boxcla:

boxsdk/object/watermark.py
```diff
@@ -0,0 +1,9 @@
+# coding: utf-8
+from __future__ import unicode_literals
+
+
+from .base_endpoint import BaseEndpoint
+
+
+class Watermark(BaseEndpoint):
```
mattwiller: This class doesn't look like it gets used anywhere.  Should we potentially make this just an API JSON object and use it to hold the watermark info instead of returning a `dict` in the item methods?

test/unit/object/conftest.py
```diff
@@ -96,6 +97,11 @@ def test_search(mock_box_session):
     return Search(mock_box_session)
 
 
+@pytest.fixture()
+def test_watermark(mock_box_session):
+    return Watermark(mock_box_session)
```
mattwiller: It doesn't look like you're using this fixture anywhere

boxsdk/object/file.py
```diff
@@ -340,3 +340,47 @@ def add_comment(self, message):
             object_id=response['id'],
             response_object=response,
         )
+
+    def get_watermark(self):
```
mattwiller: We should be consistent between this and something like `collaborate` with regards to the presence of the methods on `web_link` objects.  Any thoughts on which you would prefer: either having methods exist that don't work on certain object types, or duplicating code across multiple leaf types?

boxsdk/object/item.py
```diff
@@ -358,6 +358,52 @@ def metadata(self, scope='global', template='properties'):
...
+        """
+        Return the watermark info for a Box file
+
+        :return:
+            Watermark object.
+        :rtype:
+            :class:`Watermark`
+        """
+        url = self.get_url('watermark')
+        box_response = self._session.get(url, expect_json_response=False)
```
mattwiller: Why is this flag set?  It doesn't seem to make sense, since immediately below you call `.json()`

docs/watermarking.md
```diff
@@ -0,0 +1,54 @@
+Watermarking
+============
+
+The ability to watermark files and folders is represented as a sub-resource on the Files and Folders resources, respectively. You can think of the sub-resource as a "label" marking whether the file or folder is watermarked or not. If you apply a watermark label to a folder, then all files inside of it will be protected by the watermark (e.g. previews will be watermarked). However, those files' watermark sub-resource is independent from the folder that got watermarked. This allows you to watermark files and folders independently.
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
```
mattwiller: You can pass the `--notitle` flag to `doctoc` to get rid of this

docs/watermarking.md
```diff
@@ -0,0 +1,54 @@
...
+
+- [Watermarking](#watermarking)
```
mattwiller: You should re-run `doctoc` — this top-level Watermark link isn't really necessary

PR #326, opened 2018-08-20
Group v2

carycheng:


boxcla:

boxsdk/object/group.py
```diff
@@ -14,14 +14,14 @@ class Group(BaseObject):
     _item_type = 'group'
 
     @api_call
-    def get_memberships(self, offset=0, limit=None, fields=None):
+    def get_memberships(self, limit=None, offset=0, fields=None):
```
mattwiller: This looks like a breaking change; make sure to document it in the Changelog.  Also, consider giving `offset` a default value of `None` to align with the other methods

PR #328, opened 2018-08-21
Retention Policies v2

carycheng:


boxcla:

boxsdk/object/retention_policy.py
```diff
@@ -0,0 +1,78 @@
...
+            object_id=response['id'],
+            response_object=response,
+        )
+
+    def assignments(self, assignment_type=None, limit=None, marker=None, fields=None):
+        """Get the assignments for the retention policy.
+        :param limit:
+            The maximum number of items to return.
+        :type limit:
+            `int`
```
mattwiller: `int` or None

boxsdk/object/retention_policy.py
```diff
@@ -0,0 +1,78 @@
...
+        :param marker:
+            The position marker at which to begin the response.
+        :type marker:
+            `unicode`
```
mattwiller: `unicode` or None

boxsdk/object/retention_policy.py
```diff
@@ -0,0 +1,78 @@
...
+            List of fields to request.
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            An iterable of assignments in the retention policy.
+        :rtype:
+            `Iterable` of :class:`RetentionPolicyAssignment`
+        """
+        additional_params = {
+            'type': assignment_type,
```
mattwiller: May want to conditionally add this if not `None`

docs/retention_policy.md
```diff
@@ -0,0 +1,101 @@
+Retention Policies
+==================
+
+A retention policy blocks permanent deletion of content for a specified amount of time. Admins can create retention policies and then later assign them to specific folders or their entire enterprise.
+
+Create Retention Policy
+-----------------------
+
+You can call `client.create_retention_policy(policy_name, disposition_action)` to let you create a new indefinite retention policy.
```
mattwiller: These method signatures should have all parameters (including optional ones with default values) displayed

test/unit/object/test_retention_policy.py
```diff
@@ -0,0 +1,79 @@
...
+        'policy_name': new_policy_name,
+    })
+    data = {
+        'policy_name': new_policy_name,
+    }
+    mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(data), headers=None, params=None)
+    assert isinstance(retention_policy, RetentionPolicy)
+
+
+def test_assign(test_retention_policy, test_folder, mock_box_session):
```
mattwiller: We should consider parametrizing this test to encompass all the possible types that the policy can be assigned to

boxsdk/client/client.py
```diff
@@ -615,7 +611,7 @@ def create_retention_policy(
         """
         url = self.get_url('retention_policies')
         retention_attributes = {'policy_name': policy_name}
-        if retention_length == -1 or retention_length is None:
+        if retention_length == float('inf') or retention_length is None:
```
mattwiller: I think if we're going to go this route, we should probably make `retention_length` a required parameter (and accept any `int` or `float('inf')`)

boxsdk/client/client.py
```diff
@@ -631,7 +627,7 @@ def create_retention_policy(
             retention_attributes['custom_notification_recipients'] = custom_notification_recipients
         box_response = self._session.post(url, data=json.dumps(retention_attributes))
         response = box_response.json()
-        return self.translator.translate('retention_policy')(session=self._session, object_id=response['id'])
+        return self.translator.translate('retention_policy')(session=self._session, object_id=response['id'], response_object=response)
```
mattwiller: Should use `response['type']` instead of hardcoding the type here

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
             response_object=response,
         )
 
+    def retention_policy(self, retention_id):
+        """
+        Initialize a :class:`RetentionPolicy` object, whose box id is retention_id.
```
mattwiller: Should have a blank line between the description and the first param

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+         :param retention_id:
```
mattwiller: It looks like there's an extra space before `:param:` here

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+            A :class:`RetentionPolicy` object with the given entry ID.
+        :rtype:
+            :class:`RetentionPolicy`
+        """
+        return self.translator.translate('retention_policy')(session=self._session, object_id=retention_id)
+
+    def retention_policy_assignment(self, assignment_id):
+        """
+        Initialize a :class:`RetentionPolicyAssignment` object, whose box id is assignment_id.
+         :param assignment_id:
```
mattwiller: Blank line/extra space — same as above

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+        :type custom_notification_recipients:
+            `list` or `User` objects
+        :return:
+            The newly created Retention Policy
+        :rtype:
+            :class:`RetentionPolicy`
+        """
+        url = self.get_url('retention_policies')
+        retention_attributes = {'policy_name': policy_name}
+        if retention_length == float('inf') or retention_length is None:
```
mattwiller: I think if we're going to go this route, we should make `retention_length` a required parameter and just accept any `int` value or `float('inf')`

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+            retention_attributes['policy_type'] = 'finite'
+            retention_attributes['retention_length'] = retention_length
+            retention_attributes['disposition_action'] = disposition_action
+
+        if can_owner_extend_retention is not None:
+            retention_attributes['can_owner_extend_retention'] = can_owner_extend_retention
+        if are_owners_notified is not None:
+            retention_attributes['are_owners_notified'] = are_owners_notified
+        if custom_notification_recipients is not None:
+            retention_attributes['custom_notification_recipients'] = custom_notification_recipients
```
mattwiller: Does this work correctly when the user passes in the type described above (a `list` of `User`s)?

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+        box_response = self._session.post(url, data=json.dumps(retention_attributes))
+        response = box_response.json()
+        return self.translator.translate('retention_policy')(session=self._session, object_id=response['id'], response_object=response)
+
+    def retention_policies(
+            self,
+            policy_name=None,
+            policy_type=None,
+            created_by_user_id=None,
```
mattwiller: For parity with the rest of the SDK methods, this should probably be `created_by` and take a full `User` object

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+            limit=None,
+            marker=None,
+            fields=None,
+    ):
+        """
+        Get the entries in the retention policy using marker-based paging.
+         :param policy_name:
```
mattwiller: Blank line/extra space like above

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+            policy_name,
+            disposition_action,
+            can_owner_extend_retention=None,
+            are_owners_notified=None,
+            retention_length=None,
+            custom_notification_recipients=None,
+    ):
+        """
+        Create a retention policy for the given enterprise.
+         :param policy_name:
```
mattwiller: Blank line/extra space

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+        :type created_by_user_id:
+            `unicode` or None
+        :param limit:
+            The maximum number of entries to return per page. If not specified, then will use the server-side default.
+        :type limit:
+            `int` or None
+        :param marker:
+            The paging marker to start paging from
+        :type marker:
+            `str` or None
```
mattwiller: `unicode` or None

boxsdk/client/client.py
```diff
@@ -539,6 +539,151 @@ def create_group(self, name):
...
+        :param fields:
+            List of fields to request
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            An iterator of the entries in the retention policy
```
mattwiller: Missing an `:rtype:` here

boxsdk/object/retention_policy.py
```diff
@@ -0,0 +1,88 @@
...
+            session=self._session,
+            object_id=response['id'],
+            response_object=response,
+        )
+
+    def assignments(self, assignment_type=None, limit=None, marker=None, fields=None):
+        """Get the assignments for the retention policy.
+
+        :param assignment_type:
+            The type of retention policy assignment to retrieve.
```
mattwiller: Might be helpful to provide some examples of what goes here (e.g. `folder`, `enterprise`)

boxsdk/client/client.py
```diff
@@ -539,6 +539,251 @@ def create_group(self, name):
...
+        """
+        Create a retention policy for the given enterprise.
+
+        :param policy_name:
+            The name of the retention policy.
+        :type policy_name:
+            `unicode`
+        :param retention_length:
+            The amount of time in days to apply the retention policy to the selected content.
+            Do not specify for `indefinite` policies, only for `finite` policies.
```
mattwiller: This description is no longer accurate; the `retention_length` should be set to `float('inf')` for indefinite policies

boxsdk/object/retention_policy.py
```diff
@@ -0,0 +1,88 @@
...
+        :type assignment_type:
+            `unicode` or None. Can be set to 'folder', 'enterprise', or 'metadata_template'.
```
mattwiller: I think the examples belong in the `param` section instead of `type`

boxsdk/client/client.py
```diff
@@ -539,6 +539,251 @@ def create_group(self, name):
...
+        :type custom_notification_recipients:
+            `list` or `User` objects
+        :return:
+            The newly created Retention Policy
+        :rtype:
+            :class:`RetentionPolicy`
+        """
+        url = self.get_url('retention_policies')
+        retention_attributes = {'policy_name': policy_name}
+        if retention_length == float('inf') or retention_length == -1:
```
mattwiller: We probably don't want to allow both `float('inf')` and `-1` here — the infinity float pretty accurately describes what's going on, whereas -1 is kind of a kludge that you might have to use in strictly typed languages where this is strictly an `int` (but is unnecessary here).

boxsdk/client/client.py
```diff
@@ -539,6 +539,251 @@ def create_group(self, name):
...
+            retention_attributes['custom_notification_recipients'] = custom_notification_recipients
+        box_response = self._session.post(url, data=json.dumps(retention_attributes))
+        response = box_response.json()
+        return self.translator.translate(response['type'])(session=self._session, object_id=response['id'], response_object=response)
+
+    def get_retention_policies(
+            self,
+            policy_name=None,
+            policy_type=None,
+            created_by_user_id=None,
```
mattwiller: This should take a `User` object instead of an ID, for consistency with the rest of the SDK

boxsdk/client/client.py
```diff
@@ -539,6 +539,251 @@ def create_group(self, name):
...
+            marker=marker,
+            fields=fields,
+            return_full_pages=False,
+        )
+
+    def get_file_version_retentions(
+            self,
+            file_id=None,
+            file_version_id=None,
+            policy_id=None,
```
mattwiller: All of these params should take objects (e.g. `File` or `RetentionPolicy`) for consistency with the rest of the SDK

boxsdk/client/client.py
```diff
@@ -539,6 +539,251 @@ def create_group(self, name):
...
+        :return:
+            The newly created Retention Policy
+        :rtype:
+            :class:`RetentionPolicy`
+        """
+        url = self.get_url('retention_policies')
+        retention_attributes = {'policy_name': policy_name}
+        if retention_length == float('inf') or retention_length == -1:
+            retention_attributes['policy_type'] = 'indefinite'
+            retention_attributes['disposition_action'] = 'remove_retention'
```
mattwiller: We should still pass the `disposition_action` the user passed in here instead of silently overriding it.

test/unit/client/test_client.py
```diff
@@ -481,6 +483,94 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+        'policy_name': policy_name,
+    }
+    mock_box_session.post.return_value.json.return_value = mock_policy
+    policy = mock_client.create_retention_policy(policy_name, disposition_action, float('inf'))
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_data))
+    assert policy.object_id == mock_policy['id']
+    assert policy.object_type == mock_policy['type']
+    assert policy.policy_name == policy_name
+    assert isinstance(policy, RetentionPolicy)
+
```
mattwiller: You're missing tests for a lot of the optional values in these methods; those branches still need to be covered

test/unit/client/test_client.py
```diff
@@ -481,6 +483,94 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+    mock_retention = {
+        'type': 'file_version_retention',
+        'id': '12345',
+    }
+    mock_box_session.get.return_value.json.return_value = {
+        'limit': 100,
+        'entries': [mock_retention],
+        'next_marker': 'testMarker',
+    }
+    retentions = mock_client.get_file_version_retentions()
```
mattwiller: This method in particular has a bunch of optional params, none of which are covered by this test.  All of those branches could be working incorrectly and we'd never know

test/unit/object/test_retention_policy.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_get(test_retention_policy, mock_box_session):
+    expected_url = '{0}/retention_policies/{1}'.format(API.BASE_API_URL, test_retention_policy.object_id)
+    mock_box_session.get.return_value.json.return_value = {
+        'type': 'retention_policy',
+        'id': test_retention_policy.object_id,
+    }
+    retention_policy = test_retention_policy.get()
+    mock_box_session.get.assert_called_once_with(expected_url, headers=None, params=None)
+    assert isinstance(retention_policy, RetentionPolicy)
```
mattwiller: You should add more data to the mock API response and assert about it here; this method could be calling the translator incorrectly and not attaching the response object but this test wouldn't catch it

test/unit/object/test_retention_policy.py
```diff
@@ -0,0 +1,80 @@
...
+        'id': test_retention_policy.object_id,
+    }
+    retention_policy = test_retention_policy.update_info({
+        'policy_name': new_policy_name,
+    })
+    data = {
+        'policy_name': new_policy_name,
+    }
+    mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(data), headers=None, params=None)
+    assert isinstance(retention_policy, RetentionPolicy)
```
mattwiller: Ditto here; this should assert about the response object being accessible through the smart object

test/unit/object/test_retention_policy_assignment.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_get_assignment(test_retention_policy_assignment, mock_box_session):
+    expected_url = '{0}/retention_policy_assignments/{1}'.format(API.BASE_API_URL, test_retention_policy_assignment.object_id)
+    mock_box_session.get.return_value.json.return_value = {
+        'type': 'retention_policy_assignment',
+        'id': test_retention_policy_assignment.object_id,
+    }
+    retention_policy_assignment = test_retention_policy_assignment.get()
+    mock_box_session.get.assert_called_once_with(expected_url, headers=None, params=None)
+    assert isinstance(retention_policy_assignment, RetentionPolicyAssignment)
```
mattwiller: This should also assert about more data to ensure that the full response object is accessible

boxsdk/client/client.py
mattwiller: This doesn't look like it works as documented; a list of `User` objects would need to be formatted for the API

docs/retention_policy.md
```diff
@@ -0,0 +1,137 @@
...
+policy_id = '1234'
+folder = client.folder('1111')
+retention_policy = client.retention_policy(policy_id)
+policy_assignment = retention_policy.assign(folder)
+```
+
+Get Retention Policy Assignment
+-------------------------------
+
+Calling `retention_polict_assignment.get(fields=None, headers=None)` will return the retention policy assignment object.
```
mattwiller: [typo] "retention_polict_assignment"

test/unit/client/test_client.py
```diff
@@ -481,6 +501,143 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+        'can_owner_extend_retention': False,
+        'are_owners_notified': False,
+    }
+    mock_box_session.post.return_value.json.return_value = mock_policy
+    policy = mock_client.create_retention_policy(
+        policy_name=policy_name,
+        disposition_action=disposition_action,
+        retention_length=5,
+        can_owner_extend_retention=True,
+        are_owners_notified=False,
```
mattwiller: This should also test `custom_notification_recipients`

PR #330, opened 2018-08-21
Terms of service v2

carycheng:


boxcla:

docs/terms_of_service.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Accept or Decline a Terms of Service for New User
+-------------------------------------------------
+
+For new users you can accept or decline a terms of service by calling `terms_of_service.create_user_status(is_accepted, user=None)`.
+
+```python
+user = client.user('1234')
+user_status = client.create_user_status(True, user)
```
mattwiller: This example doesn't line up with the signature above — which class is this method on?

boxsdk/object/terms_of_service.py
```diff
@@ -0,0 +1,92 @@
...
+            },
+            'is_accepted': is_accepted,
+        }
+        if user is not None:
+            user_json = {
+                'type': user.object_type,
+                'id': user.object_id,
+            }
+            body['user'] = user_json
+        box_response = self._session.post(url, data=json.dumps(body))
```
mattwiller: This method handles the POST case, and there's one on the User Status class that handles PUT; can we also have one that handles both (similar to the Node SDK)?

docs/terms_of_service.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+You can only create a new user status on a terms of service if the user has never accepted/declined a terms of service.
+If they have then you will need to make the update call.
+
+Accept or Decline a Terms of Service for Existing User
+------------------------------------------------------
+
+For an existing user you can accept or decline a terms of service by calling `terms_of_service_user_status.accept()` to `accept` or  `terms_of_service_user_status.reject()` to `reject`
```
mattwiller: `accept` and `reject` don't need to be in backticks here, since they're not really code

docs/terms_of_service.md
```diff
@@ -0,0 +1,90 @@
...
+
+Get User Status for a Terms of Service
+-------------------------------------
+
+You can retrieve an iterable of terms of service status for a user by calling
+`terms_of_service.get_user_status(user_id=None, limit=None, fields=None)`.
+
+```python
+user_statuses = client.terms_of_service('1234').get_user_status()
+for status in user_status:
```
mattwiller: This example is incorrect; this method doesn't return an iterator

test/unit/client/test_client.py
```diff
@@ -387,6 +424,31 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+    value = json.dumps({
+        'status': 'enabled',
+        'tos_type': 'external',
+        'text': 'This is a test text',
+    })
+    mock_box_session.post.return_value = terms_of_service_response
+    new_terms_of_service = mock_client.create_a_terms_of_service('enabled', 'external', 'This is a test text')
+    assert len(mock_box_session.post.call_args_list) == 1
+    assert mock_box_session.post.call_args[0] == ("{0}/terms_of_services".format(API.BASE_API_URL),)
+    assert mock_box_session.post.call_args[1] == {'data': value}
```
mattwiller: This can probably all be simplified down to `mock_box_session.post.assert_called_once_with()`

test/unit/client/test_client.py
```diff
@@ -387,6 +424,31 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
     assert new_user.name == test_user_name
 
 
+def test_terms_of_service_returns_the_correct_tos_object(mock_client, mock_box_session, terms_of_service_response):
```
mattwiller: The test name should match up with the method under test (`create_terms_of_service` here)

test/unit/client/test_client.py
```diff
@@ -387,6 +424,31 @@ def test_create_enterprise_user_returns_the_correct_user_object(mock_client, moc
...
+    assert mock_box_session.post.call_args[0] == ("{0}/terms_of_services".format(API.BASE_API_URL),)
+    assert mock_box_session.post.call_args[1] == {'data': value}
+    assert isinstance(new_terms_of_service, TermsOfService)
+
+
+def test_get_all_terms_of_services(mock_client, mock_box_session, terms_of_services_response, tos_id_1, tos_id_2):
+    # pylint:disable=redefined-outer-name
+    mock_box_session.get.return_value = terms_of_services_response
+    services = mock_client.terms_of_services()
+    for service, expected_id in zip(services, [tos_id_1, tos_id_2]):
```
mattwiller: This test should assert about the API calls being made

test/unit/object/test_terms_of_service.py
```diff
@@ -0,0 +1,78 @@
...
+    })
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'terms_of_service_user_status',
+        'id': test_terms_of_service_user_status.object_id,
+        'created_at': created_at,
+    }
+    new_terms_of_service_user_status = test_terms_of_service.create_user_status(True, mock_user)
+    assert len(mock_box_session.post.call_args_list) == 1
+    assert mock_box_session.post.call_args[0] == ("{0}/terms_of_service_user_statuses".format(API.BASE_API_URL),)
+    assert mock_box_session.post.call_args[1] == {'data': value}
```
mattwiller: Replace with `mock_box_session.post_assert_called_once_with()`

test/unit/object/test_terms_of_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_get_user_status(test_terms_of_service, test_terms_of_service_user_status, mock_box_session):
+    created_at = '2016-05-18T17:38:03-07:00'
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'terms_of_service_user_status',
+        'id': test_terms_of_service_user_status.object_id,
+        'created_at': created_at,
+    }
+    new_terms_of_service_user_status = test_terms_of_service.get_user_status('42')
```
mattwiller: This test doesn't assert that the user ID gets passed through; you should ensure that the query params are being validated

test/unit/object/test_terms_of_service_user_status.py
```diff
@@ -0,0 +1,45 @@
...
+def test_accept(test_terms_of_service_user_status, mock_box_session):
+    expected_url = '{0}/terms_of_service_user_statuses/{1}'.format(API.BASE_API_URL, test_terms_of_service_user_status.object_id)
+    mock_box_session.put.return_value.json.return_value = {
+        'type': 'terms_of_service_user_status',
+        'id': test_terms_of_service_user_status.object_id,
+    }
+    data = {'is_accepted': True}
+    terms_of_service_user_status = test_terms_of_service_user_status.accept()
+    mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(data), headers=None, params=None)
+    assert isinstance(terms_of_service_user_status, TermsOfServiceUserStatus)
```
mattwiller: Should also assert about the ID of the user status here

test/unit/object/test_terms_of_service_user_status.py
```diff
@@ -0,0 +1,45 @@
...
+def test_reject(test_terms_of_service_user_status, mock_box_session):
+    expected_url = '{0}/terms_of_service_user_statuses/{1}'.format(API.BASE_API_URL, test_terms_of_service_user_status.object_id)
+    mock_box_session.put.return_value.json.return_value = {
+        'type': 'terms_of_service_user_status',
+        'id': test_terms_of_service_user_status.object_id,
+    }
+    data = {'is_accepted': False}
+    terms_of_service_user_status = test_terms_of_service_user_status.reject()
+    mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(data), headers=None, params=None)
+    assert isinstance(terms_of_service_user_status, TermsOfServiceUserStatus)
```
mattwiller: Should also assert about the ID of the user status here

test/unit/object/test_terms_of_service.py
```diff
@@ -0,0 +1,116 @@
...
+        'created_at': created_at,
+    }
+    new_terms_of_service_user_status = test_terms_of_service.accept(mock_user)
+    mock_box_session.post.assert_called_once_with(expected_url, data=value)
+    assert isinstance(new_terms_of_service_user_status, TermsOfServiceUserStatus)
+    assert new_terms_of_service_user_status.id == test_terms_of_service_user_status.object_id
+    assert new_terms_of_service_user_status.type == test_terms_of_service_user_status.object_type
+
+
+def test_reject_terms_of_service(test_terms_of_service, test_terms_of_service_user_status, mock_user, mock_box_session):
```
mattwiller: At least one test needs to cover the case where the POST call results in a 409 Conflict

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+        :return:
+            A :class:`TermsOfServiceUserStatus` object with the given terms of service user status id.
+        :rtype:
+            :class:`TermsOfServiceUserStatus`
+        """
+        return self.translator.get('terms_of_service_user_status')(session=self._session, object_id=tos_user_status_id)
+
+    def get_terms_of_services(self, tos_type=None, limit=None, marker=None, fields=None):
+        """
+        Get the entries in the legal hold policy assignment using limit-offset paging.
```
mattwiller: Description needs to be updated

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+
+        :param tos_type:
+            Can be set to `managed` or `external` for the type of terms of service.
+        :type tos_type:
+            `unicode`
```
mattwiller:
```suggestion
            :class:`TermsOfServiceType`
```

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+        :param: limit
+            The maximum number of items to return. If limit is set to None, then the default
+            limit (returned by Box in the response) is used. See https://developer.box.com/reference#get-recent-items
```
mattwiller: This link should probably be removed

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+            for default.
+        :type: limit
+            `int` or None
+        :param marker:
```
mattwiller: This method can only return two total items today; I don't think the user could ever need to pass in a `marker`

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+        :type marker:
+            `unicode` or None
+        :param fields:
+            List of fields to request
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            An iterator of the entries in the terms of service
+        :rtype:
+            :class:`MarkerBasedObjectCollection`
```
mattwiller:
```suggestion
            :class:`BoxObjectCollection`
```

boxsdk/client/client.py
```diff
@@ -742,6 +742,76 @@ def create_group(
...
+        """
+        additional_params = {}
+        if tos_type is not None:
+            additional_params['tos_type'] = tos_type
```
mattwiller: If you want, you could make this more succinct by doing something like this:

```python
additional_params = {'tos_type': tos_type} if tos_type is not None else None
```

boxsdk/client/client.py
```diff
@@ -940,6 +1010,41 @@ def get_retention_policies(
...
+        Create a terms of service.
+
+        :param status:
+            The status of the terms of service.
+        :type status:
+            `unicode`
+        :param tos_type:
+            The type of the terms of service. Can be set to `managed` or `external`.
+        :type tos_type:
+            `unicode`
```
mattwiller:
```suggestion
            :class:`TermsOfServiceType`
```

boxsdk/client/client.py
```diff
@@ -940,6 +1010,41 @@ def get_retention_policies(
...
         )
 
+    def create_terms_of_service(self, status, tos_type, text):
+        """
+        Create a terms of service.
+
+        :param status:
+            The status of the terms of service.
+        :type status:
+            `unicode`
```
mattwiller:
```suggestion
            :class:`TermsOfServiceStatus`
```

boxsdk/object/terms_of_service.py
```diff
@@ -0,0 +1,132 @@
...
+                'id': self.object_id,
+            },
+            'is_accepted': is_accepted,
+        }
+        if user is not None:
+            user_json = {
+                'type': user.object_type,
+                'id': user.object_id,
+            }
+            body['user'] = user_json
```
mattwiller: You don't really need a temporary variable here; you could just assign the `dict` directly to `body['user']`

boxsdk/object/terms_of_service.py
```diff
@@ -0,0 +1,132 @@
...
+        response = None
```
mattwiller: You don't need to predefine this variable here since it's not used outside of the `try` block below

docs/terms_of_service.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+Accept or Decline a Terms of Service for Existing User
+------------------------------------------------------
+
+For an existing user you can accept or decline a terms of service by calling `terms_of_service.set_user_status(is_accepted, user=None)` to accept or to reject.
+
+```python
+user = client.user('22222')
+user_status = client.terms_of_service('11111').set_user_status(is_accepted=True, user=user)
```
mattwiller: This isn't really the best example for this; if you know the user has already accepted or declined the ToS you should not use the convenience method this way since it will make an extra API call

docs/terms_of_service.md
```diff
@@ -0,0 +1,84 @@
...
+
+Get User Status for a Terms of Service
+-------------------------------------
+
+You can retrieve an iterable of terms of service status for a user by calling
+`terms_of_service.get_user_status(user_id=None)`.
+
+```python
+user_id = '5678'
+user_status = client.terms_of_service('1234').get_user_status(user_id)
```
mattwiller: This method takes a `User`, not a user ID

boxsdk/client/client.py
```diff
@@ -940,6 +1005,41 @@ def get_retention_policies(
...
+        """
+        url = self.get_url('terms_of_services')
+        body = {
+            'status': status,
+            'tos_type': tos_type,
+            'text': text
+        }
+        box_response = self._session.post(url, data=json.dumps(body))
+        response = box_response.json()
+        return self.translator.translate(response['type'])(
```
mattwiller: This will need to be updated now that the new translator is merged

boxsdk/object/terms_of_service.py
```diff
@@ -0,0 +1,130 @@
...
+        url = self._session.get_url('terms_of_service_user_statuses')
+        additional_params = {
+            'tos_id': self.object_id,
+        }
+        if user is not None:
+            additional_params['user_id'] = user.object_id
+        box_response = self._session.get(url, params=additional_params)
+        response_object = box_response.json()
+        response = response_object['entries'][0]
+        return self.translator.translate(response['type'])(
```
mattwiller: This will need to be updated now that the new translator is merged

boxsdk/object/terms_of_service.py
```diff
@@ -0,0 +1,130 @@
...
+        if user is not None:
+            body['user'] = {
+                'type': user.object_type,
+                'id': user.object_id,
+            }
+        translated_response = None
+        try:
+            box_response = self._session.post(url, data=json.dumps(body))
+            response = box_response.json()
+            translated_response = self.translator.translate(response['type'])(
```
mattwiller: This will need to be updated now that the new translator is merged

docs/terms_of_service.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+Accept or Decline a Terms of Service for Existing User
+------------------------------------------------------
+
+For new users you can accept or decline a terms of service by calling `terms_of_service.accept(user=None)` or `terms_of_service.reject(user=None)`.
+
+```python
+user = client.user('1234')
+user_status = client.terms_of_service('11111').accept(user)
```
mattwiller: I don't think this example is quite right, since `accept()` just calls `set_user_status()` under the hood.  This should bypass trying to create a new status and just update the existing one for the user, based on the title of this section.

docs/terms_of_service.md
```diff
@@ -0,0 +1,99 @@
...
+```python
+terms_of_services = client.get_terms_of_services()
+for terms_of_service in terms_of_services:
+    # Do something
+```
+
+Accept or Decline a Terms of Service for Existing User
+------------------------------------------------------
+
+For new users you can accept or decline a terms of service by calling `terms_of_service.accept(user=None)` or `terms_of_service.reject(user=None)`.
```
mattwiller: This says "for Existing User" in the section title, but this says it's for "new users" — this could use some more explanation/clarification.

docs/terms_of_service.md
```diff
@@ -0,0 +1,99 @@
...
+user = client.user('22222')
+user_status = client.terms_of_service('11111').set_user_status(is_accepted=True, user=user)
+```
+
+It is important to note that regardless of whether the user has taken action on this terms of service. This will create and update the user status on the terms of service.
+
+Get User Status for a Terms of Service
+-------------------------------------
+
+You can retrieve an iterable of terms of service status for a user by calling
```
mattwiller: I don't think this returns an iterable

PR #334, opened 2018-08-22
User v2

carycheng:


boxsdk/object/user.py
```diff
@@ -1,16 +1,128 @@
...
+        :type email_alias:
+            :class:`EmailAlias`
+        :returns:
+            Whether the removal succeeded.
+        :rtype:
+            `bool`
+        """
+        url = self.get_url('email_aliases', email_alias.object_id)
+        response = self._session.delete(url, expect_json_response=False)
+        return response.ok
```
jmoldow: This should just be `return email_alias.delete()`

Is there any reason this method needs to exist?

boxsdk/object/user.py
```diff
@@ -43,3 +155,28 @@ def get_group_memberships(self, limit=None, offset=None, fields=None):
...
+            Whether the deletion succeeded
+        :rtype:
+            `bool`
+        """
+        url = self.get_url()
+        params = {
+            'notify': notify,
+            'force': force,
+        }
+        response = self._session.delete(url, params=params, expect_json_response=False)
```
jmoldow: This should be `return super(User, self).delete(params=params)`

boxsdk/object/user.py
jmoldow: Oh, I didn't notice that. I guess this is fine for now then.

Or we can move this implementation into an override of `delete()` on EmailAlias, and have it enforce that you give a `user` input.

PR #336, opened 2018-08-22
Tasks v2

carycheng:


boxcla:

boxsdk/client/client.py
```diff
@@ -585,6 +585,36 @@ def create_group(
...
+        :param task_id:
+            The box ID of the :class:`Task` object.
+        :type task_id:
+            `unicode`
+        :return:
+            A :class:`Task` object with the given entry ID.
+        :rtype:
+            :class:`Task`
+        """
+        return self.translator.translate('task')(session=self._session, object_id=task_id)
```
mattwiller: It would be nice to switch these sorts of usages over to use `translator.get()` instead — it does the same thing, and will be forwards-compatible with the nested object translation work I'm doing

boxsdk/client/client.py
```diff
@@ -585,6 +585,36 @@ def create_group(
...
+        :param assignment_id:
+            The box ID of the :class:`TaskAssignment` object.
+        :type assignment_id:
+            `unicode`
+        :return:
+            A :class:`TaskAssignment` object with the given entry ID.
+        :rtype:
+            :class:`TaskAssignment`
+        """
+        return self.translator.translate('task_assignment')(session=self._session, object_id=assignment_id)
```
mattwiller: Use `translator.get()` — see above

boxsdk/object/file.py
```diff
@@ -340,3 +341,70 @@ def add_comment(self, message):
...
+            task_attributes['due_at'] = due_at
+        box_response = self._session.post(url, data=json.dumps(task_attributes))
+        response = box_response.json()
+        return self.translator.translate(response['type'])(
+            session=self._session,
+            object_id=response['id'],
+            response_object=response,
+        )
+
+    def get_tasks(self, limit=None, marker=None, fields=None):
```
mattwiller: This API doesn't support paging; `limit` and `marker` should be removed

boxsdk/object/task.py
```diff
@@ -0,0 +1,78 @@
...
+
+from boxsdk.object.base_object import BaseObject
+from ..pagination.marker_based_object_collection import MarkerBasedObjectCollection
+
+
+class Task(BaseObject):
+    """Represents a Box task."""
+    _item_type = 'task'
+
+    def assign(self, assignee, assign_to_login=None):
```
mattwiller: This should be split into two different methods; a user can't pass both params and expect something reasonable to happen

boxsdk/object/task.py
```diff
@@ -0,0 +1,78 @@
...
+        if assign_to_login is not None:
+            body['assign_to']['login'] = assign_to_login
+        response = self._session.post(url, data=json.dumps(body)).json()
+        return self.translator.translate(response['type'])(
+            session=self._session,
+            object_id=response['id'],
+            response_object=response,
+        )
+
+    def get_assignments(self, limit=None, marker=None, fields=None):
```
mattwiller: This API doesn't support paging; `limit` and `marker` should be removed

docs/task.md
```diff
@@ -0,0 +1,107 @@
...
+```python
+message = 'Please review this'
+due_at = "2014-04-03T11:09:43-07:00"
+task = client.file('11111').create_task(message, due_at)
+```
+
+Update Task Info
+----------------
+
+To update the task information use, `task.update_info(data, params=None, headers=None, **kwargs)`
```
mattwiller: What might the user want to pass in for `params` or `headers`, and what are the `kwargs` used for?  These base class properties should either be better-documented or removed.

docs/task.md
```diff
@@ -0,0 +1,107 @@
...
+
+```python
+task_assignment_id = '2222'
+assignment_info = client.task_assignment(task_assignment_id).get()
+```
+
+Update Task Assignment
+----------------------
+
+To update information about a task assignment use, `task_assignment.update_info(data, params=None, headers=None, **kwargs)`
```
mattwiller: The base class params in `update_info()` should be removed if they don't apply

test/unit/object/test_file.py
```diff
@@ -41,6 +42,58 @@ def test_delete_file(test_file, mock_box_session, etag, if_match_header):
...
+    # pylint:disable=redefined-outer-name
+    expected_url = "{0}/tasks".format(API.BASE_API_URL)
+    expected_body = {
+        'item': {
+            'type': 'file',
+            'id': '42',
+        },
+        'action': 'review',
+        'message': 'Test Message',
+        'due_at': '2014-04-03T11:09:43-07:00',
```
mattwiller: You might want to extract these values into variables to make it easier to put them in all three pleaces (this `expected_body`, the `return_value` below, and the actual invocation of `create_task()` below that).

test/unit/object/test_file.py
```diff
@@ -41,6 +42,58 @@ def test_delete_file(test_file, mock_box_session, etag, if_match_header):
...
+        'message': 'Test Message',
+    }
+    value = json.dumps(expected_body)
+    new_task = test_file.create_task(message='Test Message', due_at='2014-04-03T11:09:43-07:00')
+    mock_box_session.post.assert_called_once_with(expected_url, data=value)
+    assert isinstance(new_task, Task)
+    assert new_task.object_type == test_task.object_type
+    assert new_task.object_id == test_task.object_id
+    assert new_task.action == 'review'
+    assert new_task.message == 'Test Message'
```
mattwiller: Please also assert about `due_at`, especially since it's an input parameter

test/unit/object/test_task.py
```diff
@@ -0,0 +1,113 @@
...
+        'id': test_task.object_id,
+        'due_at': '2014-04-03T11:09:43-07:00',
+        'action': 'review',
+        'message': 'Test Message',
+    }
+    retrieved_task = test_task.get()
+    mock_box_session.get.assert_called_once_with(expected_url, headers=None, params=None)
+    assert isinstance(retrieved_task, Task)
+    assert retrieved_task.object_type == test_task.object_type
+    assert retrieved_task.object_id == test_task.object_id
```
mattwiller: Need to assert about the rest of the object fields here

test/unit/object/test_task.py
```diff
@@ -0,0 +1,113 @@
...
+    }
+    mock_box_session.post.return_value.json.return_value = {
+        'type': 'task_assignment',
+        'id': '42',
+    }
+    new_legal_hold_assignment = test_task.assign(assignee=mock_user, assign_to_login='test_user@example.com')
+    mock_box_session.post.assert_called_once_with(expected_url, data=json.dumps(expected_body))
+    assert isinstance(new_legal_hold_assignment, TaskAssignment)
+    assert new_legal_hold_assignment.object_type == 'task_assignment'
+    assert new_legal_hold_assignment.object_id == '42'
```
mattwiller: Two things: these variable names need to be updated so they don't refer to legal holds, and you need to assert about the rest of the body to verify that the object translation worked correctly

test/unit/object/test_task_assignment.py
```diff
@@ -0,0 +1,72 @@
...
+def test_delete_policy_return_the_correct_response(
+        test_task_assignment,
+        mock_box_session,
+        delete_task_assignment_response,
+):
+    # pylint:disable=redefined-outer-name
+    mock_box_session.delete.return_value = delete_task_assignment_response
+    response = test_task_assignment.delete()
+    # pylint:disable=protected-access
+    expected_url = test_task_assignment.get_url()
```
mattwiller: This URL should be hardcoded to ensure the test actually checks that the implementation is correct

test/unit/object/test_task.py
```diff
@@ -0,0 +1,113 @@
...
+def test_delete_policy_return_the_correct_response(
+        test_task,
+        mock_box_session,
+        delete_task_response,
+):
+    # pylint:disable=redefined-outer-name
+    mock_box_session.delete.return_value = delete_task_response
+    response = test_task.delete()
+    # pylint:disable=protected-access
+    expected_url = test_task.get_url()
```
mattwiller: This URL should be hardcoded to ensure the test actually checks that the implementation is correct

test/unit/object/test_task_assignment.py
```diff
@@ -0,0 +1,72 @@
...
+        'message': new_message,
+    }
+    expected_body = {
+        'message': new_message,
+        'resolution_state': resolution_state,
+    }
+    updated_task_assignment = test_task_assignment.update_info({'message': new_message, 'resolution_state': resolution_state})
+    mock_box_session.put.assert_called_once_with(expected_url, data=json.dumps(expected_body), headers=None, params=None)
+    assert isinstance(updated_task_assignment, TaskAssignment)
+    assert updated_task_assignment.message == new_message
```
mattwiller: Should also assert against the `resolution_state`, since it's an input parameter

test/unit/object/test_task_assignment.py
mattwiller: This URL should also be hardcoded

PR #337, opened 2018-08-22
root folder functionality

carycheng:


boxcla:

boxsdk/client/client.py
```diff
@@ -84,6 +84,12 @@ def folder(self, folder_id):
         """
         return self.translator.translate('folder')(session=self._session, object_id=folder_id)
 
+    def root_folder(self):
+        """
+        Returns a user's root folder object.
+        """
+        return self.translator.translate('folder')(self._session, '0')
```
Jeff-Meadows: I would just use `self.folder('0')`

PR #345, opened 2018-10-02
Run CI tests on Python 3.7

mattwiller:


boxcla:

mattwiller:
It looks like tests are broken on 3.7 due to a change in behavior around handling StopIteration inside generators.  The line where the error originates is actually [inside of mock](https://github.com/testing-cabal/mock/blob/8c19ef6c95004524cb9f0b170a1aff6f47ade764/mock/mock.py#L1115); it should ostensibly be guarding the call to `next()` with a try/except to ensure that StopIteration does not leave the stack frame (whereupon it will be converted to a RuntimeError).

jmoldow:
The problem appears to be ours.

The `StopIteration` on the line you pointed to in `mock` is legitimate. That line `result = next(effect)` is allowed / expected to raise `StopIteration`, because that's part of the interface of `next()`.

The code in that function has no requirement to catch `StopIteration`, because it isn't a generator function. In fact, it's probably quite intentional that `StopIteration` isn't caught in `_mock_call()`.

Moving up the stack, `__call__` is also not a generator function, and doesn't have to catch `StopIteration`, and again, this seems to be intentional.

The next function in the stack is our own, and indeed, we're calling the mocked `system_random.random()` inside a generator: `(ascii_alphabet[int(system_random.random() * ascii_len)] for _ in range(jti_length))`.

Previously, the `StopIteration` would get caught and kill the generator, but in Python 3.7 that is no longer legal.

It seems that our test code wasn't functioning properly, but we didn't notice because it was silently ignored in older versions of Python. The behavior change exposed our bug.

We should be mocking `system_random.random()` such that we are able to call it the number of times that we need it to, without it triggering `StopIteration`.

jmoldow:
Though I'm confused, because it looks like the data size for the test fixture should be the same as the number of times we call it in the code.

PR #350, opened 2018-10-09
Chunked uploader

carycheng:


boxcla:

boxsdk/object/file.py
```diff
@@ -36,6 +37,42 @@ def preflight_check(self, size, name=None):
             file_id=self._object_id,
         )
 
+    def create_upload_session(self, file_size, file_name=None):
+        """
+        Create a new chunked upload session for uploading a new version of the file.
+
+        :param file_size:
+            The size of the file that will be uploaded.
```
mattwiller: Might be worth noting that this is in bytes

boxsdk/object/file.py
```diff
@@ -36,6 +37,42 @@ def preflight_check(self, size, name=None):
...
+        :type file_size:
+            `int`
+        :param file_name:
+            The name of the file version that will be uploaded.
```
mattwiller: This isn't totally correct; this parameter renames the entire file on upload.

boxsdk/object/folder.py
```diff
@@ -112,6 +113,39 @@ def preflight_check(self, size, name):
...
+        url = '{0}/files/upload_sessions'.format(API.UPLOAD_URL)
+        body_params = {
+            'folder_id': self.object_id,
+            'file_size': file_size,
+            'file_name': file_name,
+        }
+        response = self._session.post(
+            url,
+            data=json.dumps(body_params),
+        ).json()
```
mattwiller: This is a bit hard to read spread across multiple lines; does it fit all on one line instead?

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,146 @@
...
+            self.get_url(),
+            headers={
+                'Content-Type': 'application/octet-stream',
+                'Digest': 'SHA={0}'.format(base64.b64encode(part_content_sha1).decode('utf-8')),
+                'Content-Range': 'bytes {0}-{1}/{2}'.format(offset, range_end, total_size),
+            },
+            data=content_stream
+        ).json()
+
+    def commit(self, parts, content_sha1):
```
mattwiller: This should also allow passing arbitrary file attributes and an ETag; see https://developer.box.com/v2.0/reference/#commit-upload

Also, `parts` can be optional — if the user doesn't pass it in you can just call the API to get the list of parts the server has and use those (if you just want to trust the server).

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,146 @@
...
+            The uploaded part.
+        :rtype:
+            `dict`
+        """
+        if part_content_sha1 is None:
+            part_content_sha1 = self._calculate_part_sha1(content_stream)
+
+        range_end = min(offset + self.part_size - 1, total_size - 1)  # pylint:disable=no-member
+
+        return self._session.put(
```
mattwiller: It might be good to return just the actual part `dict` here, to make it easier for developers to form the correct array for `commit()`

test/unit/object/test_upload_session.py
```diff
@@ -0,0 +1,116 @@
...
+    return UploadSession(mock_box_session, '11493C07ED3EABB6E59874D3A1EF3581', upload_session_response_object)
+
+
+def test_get_parts(test_upload_session, mock_box_session):
+    expected_url = '{0}/files/upload_sessions/{1}/parts'.format(API.UPLOAD_URL, test_upload_session.object_id)
+
+    parts = mock_box_session.get.return_value.json.return_value = {
+        'entries': [
+            {
+                'part': {
```
mattwiller: According to the API reference, this isn't what the response looks like.  Please align with the actual response format and ensure that this still works correctly

boxsdk/object/file.py
```diff
@@ -36,6 +37,37 @@ def preflight_check(self, size, name=None):
...
+        :rtype:
+            :class:`ChunkedUploadSession`
+        """
+        body_params = {
+            'file_id': self.object_id,
+            'file_size': file_size,
+        }
+        if file_name is not None:
+            body_params['file_name'] = file_name
+        url = self.get_url('{0}'.format('upload_sessions')).replace(API.BASE_API_URL, API.UPLOAD_URL)
```
Jeff-Meadows:
```suggestion
        url = self.get_url('{0}'.format('upload_sessions')).replace(
            self._session.api_config.BASE_API_URL,
            self._session.api_config.UPLOAD_URL,
        )
```

boxsdk/object/folder.py
```diff
@@ -112,6 +113,36 @@ def preflight_check(self, size, name):
...
+        :param file_name:
+            The new name of the file that will be uploaded.
+        :type file_name:
+            `unicode`
+        :returns:
+            A :class:`ChunkedUploadSession` object.
+        :rtype:
+            :class:`ChunkedUploadSession`
+        """
+        url = '{0}/files/upload_sessions'.format(API.UPLOAD_URL)
```
Jeff-Meadows:
```suggestion
        url = '{0}/files/upload_sessions'.format(self._session.api_config.UPLOAD_URL))
```

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,167 @@
...
+import base64
+import hashlib
+import json
+
+from .base_object import BaseObject
+from ..config import API
+from ..pagination.chunked_upload_part_limit_offset_based_object_collection import ChunkedUploadPartLimitOffsetBasedObjectCollection
+
+
+class UploadSession(BaseObject):
```
Jeff-Meadows: Naming should be consistent for the various objects. Either everything should start with `ChunkedUpload` or just `Upload` (my preference is the explicit `ChunkedUpload` but just upload could work).

But don't have an `UploadSession` that uses a class named `ChunkedUploadPartLimitOffsetBasedObjectCollection`. Also, the docstrings for creating upload sessions in `File` and `Folder` refer to this class name as `ChunkedUploadSession`, so make sure the names agree here and there.

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,167 @@
...
+        Base class override. Endpoint is a little different - it's /files/upload_sessions.
+
+        :rtype:
+            `unicode`
+        """
+        return self.session.get_url(
+            '{0}s/{1}s'.format(self._parent_item_type, self._item_type),
+            self._object_id,
+            *args
+        ).replace(API.BASE_API_URL, API.UPLOAD_URL)
```
Jeff-Meadows:
```suggestion
        ).replace(self.session.api_config.BASE_API_URL, self.session.api_config.UPLOAD_URL)
```

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,167 @@
...
+        range_end = min(offset + self.part_size - 1, total_size - 1)  # pylint:disable=no-member
+
+        return self._session.put(
+            self.get_url(),
+            headers={
+                'Content-Type': 'application/octet-stream',
+                'Digest': 'SHA={0}'.format(base64.b64encode(part_content_sha1).decode('utf-8')),
+                'Content-Range': 'bytes {0}-{1}/{2}'.format(offset, range_end, total_size),
+            },
+            data=part_bytes
```
Jeff-Meadows:
```suggestion
            data=part_bytes,
```

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+    def upload_part(self, part_bytes, offset, total_size, part_content_sha1=None):
+        """
+        Upload a part of a file.
+
+        :param part_bytes:
+            Part bytes
+        :type part_bytes:
+            `bytes`
```
Jeff-Meadows: This is a pretty unfriendly interface, since it doesn't allow streaming and requires the entire part content to stay in memory until it's completely uploaded.

I'd suggest keeping the interface as a content stream, and perhaps including a convenience method that takes `bytes`. That method could wrap them in a `BytesIO` and call the method that takes a content stream.

boxsdk/object/upload_session.py
Jeff-Meadows: That would be a nice method on a `ChunkedUploadPart` object.

boxsdk/pagination/chunked_upload_part_page.py
```diff
@@ -0,0 +1,12 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+from .page import Page
+
+
+class ChunkedUploadPartPage(Page):
```
Jeff-Meadows: document this class - what's it for?

docs/file.md
```diff
@@ -0,0 +1,88 @@
...
+  - [Upload Part](#upload-part)
+  - [Commit Upload Session](#commit-upload-session)
+  - [Abort Upload Session](#abort-upload-session)
+  - [List Upload Parts](#list-upload-parts)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+### Manual Process
+
+For more complicated upload scenarios, such as those being coordinated across
```
Jeff-Meadows: This paragraph doesn't make sense. The manual process is the only process implemented.

docs/file.md
```diff
@@ -0,0 +1,88 @@
...
+uploader, the endpoints for chunked upload operations are also exposed directly.
+
+The individual endpoint methods are detailed below:
+
+#### Create Upload Session for File Version
+
+To create an upload session for uploading a large version, use `file.create_upload_session(file_size, file_name=None)`
+
+```python
+file_size = 197520
```
Jeff-Meadows: Consider using a realistic size, since our API won't let you upload such a small file in chunks.

docs/file.md
```diff
@@ -0,0 +1,88 @@
...
+upload_session = client.folder('22222').create_upload_session(file_size=file_size, file_name=file_name)
+```
+
+#### Upload Part
+
+To upload a part of the file to this session, use `upload_session.upload_part(part_bytes, offset, total_size, part_content_sha1=None)`
+
+```python
+from io import BytesIO
+offset = 32
```
Jeff-Meadows: This example isn't likely to help a developer. The numbers don't match what the API will allow, and the code to read `chunk_size` bytes from a stream is unlikely to work for a real chunk size.

docs/file.md
```diff
@@ -0,0 +1,88 @@
...
+part_bytes = BytesIO(b'abcdefgh')
+upload_session = client.upload_session('11493C07ED3EABB6E59874D3A1EF3581')
+chunk = part_bytes.read(upload_session.part_size)
+offset = 32
+upload_part = upload_session.upload_part(chunk, offset, total_size)
```
Jeff-Meadows: What should a developer _do_ with the resulting upload part?

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,167 @@
...
+    def abort(self):
+        """
+        Abort an upload session, cancelling the upload and removing any parts that have already been uploaded.
+
+        :returns:
+            A boolean indication success of the upload abort.
+        :rtype:
+            `bool`
+        """
+        response = self._session.delete(self.get_url())
```
Jeff-Meadows:
```suggestion
        return self.delete()
```

docs/file.md
```diff
@@ -0,0 +1,88 @@
+Chunked Upload
```
Jeff-Meadows: I think this document needs some improvement. I don't think a developer would have a very good idea of how to use chunked uploads after reading it. Perhaps we could have an end-to-end example or something.

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,165 @@
...
+        :param offset:
+            The index at which to start returning items.
+        :type offset:
+            `int` or None
+        :param fields:
+            Fields to include on the returned items.
+        :type fields:
+            `Iterable` of `unicode`
+        :returns:
+            Returns a `list` of parts uploaded so far.
```
mattwiller: This returns a `BoxObjectCollection` instead of a `list` now

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,165 @@
...
+        :type offset:
+            `int`
+        :param total_size:
+            The size of the file that this part belongs to.
+        :type total_size:
+            `int`
+        :param part_content_sha1:
+            SHA-1 hash of the part's content. If not specified, this will be calculated.
+        :type part_content_sha1:
+            `unicode` or None
```
mattwiller: Is this the right type?  Below you encode the hash in a way that I think requires `bytes`

boxsdk/object/upload_session.py
```diff
@@ -0,0 +1,165 @@
...
+        body = {}
+        parts_list = []
+        if file_attributes is not None:
+            body['attributes'] = file_attributes
+        if parts is not None:
+            body['parts'] = parts
+        else:
+            for part in self.get_parts():
+                parts_list.append(part)
+            body['parts'] = parts_list
```
mattwiller: This should be simplified to a list comprehension:

```python
body['parts'] = [part for part in self.get_parts()]

File: boxsdk/object/upload_session.py [hunk 23]
```diff
@@ -0,0 +1,165 @@
...
+        :rtype:
+            :class:`File`
+        """
+        body = {}
+        parts_list = []
+        if file_attributes is not None:
+            body['attributes'] = file_attributes
+        if parts is not None:
+            body['parts'] = parts
+        else:
```
mattwiller: In your manual test we got things working by keeping track of the parts (and so using the `if` branch above).  Have you tested to verify that the `else` branch here works correctly?

carycheng: Just manually tested the else and the get_parts() works as well for the commit

PR #353, opened 2018-10-25
Metadata template endpoints

mattwiller:


boxcla:

boxsdk/util/translator.py
```diff
@@ -28,12 +28,33 @@ def _get_object_id(obj):
         `dict`
     :return:
     """
-    if obj.get('type', '') == 'event':
+    obj_type = obj.get('type')
```
Jeff-Meadows: This logic probably belongs in object classes. I know this method isn't new as of this review, but it's becoming more obvious that this is the wrong place for this logic.

mattwiller:
@Jeff-Meadows I added the field exclusion mechanism so object classes can specify which fields should not be translated.  I wasn't certain how we might want to decentralize the logic for finding an ID, so I left it in the translator for now.

boxsdk/object/metadata_template.py
```diff
@@ -221,6 +221,7 @@ class MetadataTemplate(BaseObject):
     """Represents a metadata template, which contains the the type information for associated metadata fields."""
 
     _item_type = 'metadata_template'
+    _untranslated_fields = ['fields']
```
Jeff-Meadows: tuples are preferable for readonly data like this, so I'd make it `('fields',)`

boxsdk/object/base_api_json_object.py
```diff
@@ -72,6 +72,7 @@ class BaseAPIJSONObject(object):
     # also important to add the module name to __all__ in object/__init__.py,
     # so that it will be imported and registered with the default translator.
     _item_type = None
+    _untranslated_fields = None
```
Jeff-Meadows: This could be `()` instead of `None`

boxsdk/util/translator.py
```diff
@@ -183,7 +166,14 @@ def translate(self, session, response_object):
             return response_object
 
         translated_obj = {}
+        object_type = response_object.get('type', '')
+        if object_type is not None:
+            # Parent classes have the ability to "blacklist" fields that they do not want translated
+            blacklisted_fields = self.get(object_type)._untranslated_fields or []  # pylint: disable=protected-access
```
Jeff-Meadows: Perhaps we could have a property for reading this so we don't have to access the private field here.

boxsdk/object/metadata_template.py
```diff
@@ -0,0 +1,282 @@
...
+    DATE = 'date'
+    ENUM = 'enum'
+    MULTISELECT = 'multiSelect'
+    FLOAT = 'float'
+
+
+class MetadataField(object):
+    """Represents a metadata field when creating or updating a metadata template."""
+
+    def __init__(self, field_type, display_name, key=None, options=None):
```
Jeff-Meadows: docstring

boxsdk/object/metadata_template.py
```diff
@@ -0,0 +1,282 @@
...
+        :type field_key:
+            `unicode`
+        """
+        self.ops.append({
+            'op': 'removeField',
+            'fieldKey': field_key,
+        })
+
+
+class MetadataFieldType(TextEnum):
```
Jeff-Meadows: is this only used in tests?

maybe it could be used in the docstring of the `MetadataField` constructor

boxsdk/object/metadata_template.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def __init__(self, field_type, display_name, key=None, options=None):
+        self.type = field_type
+        self.name = display_name
+        self.key = key
+        self.options = options
+
+    def json(self):
+        """
+        Returns the correct representation of the temnplate field for the API.
```
Jeff-Meadows: typo: template

boxsdk/object/metadata_template.py
```diff
@@ -0,0 +1,282 @@
...
+            values['type'] = self.type
+
+        if self.name is not None:
+            values['displayName'] = self.name
+
+        if self.key is not None:
+            values['key'] = self.key
+
+        if self.type in ['enum', 'multiSelect']:
+            values['options'] = [{'key': opt} for opt in self.options] if self.options is not None else []
```
Jeff-Meadows: could simplify to `values['options'] = [{'key': opt} for opt in self.options or ()]`

boxsdk/util/translator.py
```diff
@@ -173,8 +184,11 @@ def translate(self, session, response_object):
         # Try to translate any API object with a `type` property, except for metadata instances
         # The $type value in metadata instances isn't directly usable, so we avoid it altogether
         # NOTE: Currently, we represent metadata as just a `dict`, so there's no need to translate it anyway
+        # Metadata field objects are another issue; they contain a 'type' property that doesn't really
+        # map to a Box object.  We probably want to treat these as just `dict`s, so they're excluded here
         if 'type' in translated_obj and '$type' not in translated_obj:
-            object_class = self.get(translated_obj.get('type', ''))
+            object_type = translated_obj.get('type', '')
```
jmoldow: Wouldn't this be identical to `object_type` that you already computed above?

boxsdk/client/client.py
```diff
@@ -1300,3 +1300,127 @@ def device_pinners(self, enterprise_id, direction=None, limit=None, marker=None,
...
+        :type template_key:
+            `unicode`
+        :returns:
+            The metadata template object
+        :rtype:
+            :class:`MetadataTemplate`
+        """
+        return self.translator.get('metadata_template')(
+            session=self._session,
+            object_id='{0}/{1}'.format(scope, template_key),
```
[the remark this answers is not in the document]
jmoldow: Thinking about this more, I feel even more strongly that we should make this change.

I think the current approach is going to cause us problems, now and in the future.

You can already see one problem forming in this PR, where there's a bunch of places in the code where we need to manually concat the scope and templateKey to form the object_id.

Another place we're going to run into problems is with POST and PUT operations where the body expects metadata template. If we naively do `template.object_id`, that might be incorrect if the API expects the actual id (I know of at least one API that was supposed to be doing this in the future). If we do `template.id`, that field might not actually be available, if it wasn't constructed with one.

For simplicity, we could make it so that if we're constructed with the id, we fetch the scope and templateKey, and also vice-versa, that way we always have all of those pieces of information available.

boxsdk/client/client.py
```diff
@@ -1300,3 +1300,127 @@ def device_pinners(self, enterprise_id, direction=None, limit=None, marker=None,
...
+            url=self._session.get_url('metadata_templates', scope),
+            session=self._session,
+            limit=limit,
+            marker=marker,
+            fields=fields,
+            return_full_pages=False,
+        )
+
+    @api_call
+    def create_metadata_template(self, display_name, fields, template_key=None, hidden=False, scope='enterprise'):
```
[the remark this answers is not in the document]
mattwiller: I don't mind moving it, but I put it last since at this time only enterprise scope templates can be created.  Nobody calling this method today (and likely for the near future) will be setting that parameter, whereas they will likely want to set the others.

boxsdk/object/metadata_template.py
```diff
@@ -0,0 +1,314 @@
...
+            values['options'] = [{'key': opt} for opt in self.options or ()]
+
+        return values
+
+
+class MetadataTemplate(BaseObject):
+    """Represents a metadata template, which contains the the type information for associated metadata fields."""
+
+    _item_type = 'metadata_template'
+    _untranslated_fields = ('fields',)
```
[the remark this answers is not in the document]
mattwiller: The objects in the `fields` array have `type` property that doesn't map to a real Box object type, but to the type of field they are (e.g. 'string' or 'multiSelect').  Any recommendations on how this should be handled?

boxsdk/util/translator.py
```diff
@@ -28,9 +28,13 @@ def _get_object_id(obj):
         `dict`
     :return:
     """
-    if obj.get('type', '') == 'event':
+    obj_type = obj.get('type')
+    if obj_type == 'event':
         return obj.get('event_id', None)
```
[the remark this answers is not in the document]
mattwiller: I mostly just added it here for completeness; I can remove it if you want — shouldn't matter.

boxsdk/client/client.py
```diff
@@ -1481,3 +1481,130 @@ def device_pinners(self, enterprise_id, direction=None, limit=None, marker=None,
...
+            object_id=None,
+            response_object={
+                'type': 'metadata_template',
+                'scope': scope,
+                'templateKey': template_key,
+            },
+        )
+
+    @api_call
+    def get_metadata_template_by_id(self, template_id):
```
jmoldow: Should we get rid of the `.get()` behavior here? (and if so, rename the method)

test/unit/util/test_translator.py
```diff
@@ -195,3 +211,15 @@ def test_translate(default_translator, mock_box_session):
...
+def test_translator_translates_metadata_template(default_translator, mock_box_session, metadata_template_response):
+    metadata_template = default_translator.translate(mock_box_session, metadata_template_response)
+
+    if metadata_template_response.get('id'):
+        assert metadata_template.id == metadata_template_response['id']
+        assert metadata_template.object_id == metadata_template_response['id']
+
+    if metadata_template_response.get('scope') and metadata_template_response.get('templateKey'):
+        assert metadata_template.scope == metadata_template_response['scope']
+        assert metadata_template.template_key == metadata_template_response['templateKey']
```
jmoldow: I think you have this in a different test too, but just for completeness, can we assert that `metadata_template.object_id is None` if `metadata_template_response.get('id') is None`

PR #355, opened 2018-10-25
Accept date object or string for datetime inputs

mattwiller:


boxcla:

test/unit/object/conftest.py
```diff
@@ -322,3 +322,39 @@ def shared_link_unshared_at(request):
...
     return request.param
+
+
+@pytest.fixture()
+def test_timezone():
+
+    class LocalTimezone(tzinfo):
+
+        def utcoffset(self):
+            return timedelta(minutes=-8 * 60)
```
Jeff-Meadows: just use `hours=-8`

mattwiller:
@Jeff-Meadows Can you elaborate on why this wouldn't be a good idea?  I had noticed that the previous SDK methods (e.g. `item.create_shared_link()`) took a `date`, so I was trying to align with that.  Is that not the preferred way anymore?  Should we go the other way and have that method just take a string instead?

Jeff-Meadows:
@mattwiller The reason I'm not sure if it's a good idea is because it makes the SDK seem to have an inconsistent API. Consider a user of one of these methods - eg create legal hold policy. They give us a `date`/`datetime` into the `create_legal_hold_policy` method, and we give them back a `LegalHoldPolicy` object. If they access one of the date properties of that object, they'll get back a string.

mattwiller:
@Jeff-Meadows That makes sense to me.  Should I change the existing `item.create_shared_link()` interface to make it take a `unicode` string for consistency?

Jeff-Meadows:
@mattwiller I think we should be consistent. We may want to make it easy for a user to know how to format these things, so maybe some documentation is an order.

mattwiller:
@Jeff-Meadows Sounds good — I'll close this and make a new PR for that

PR #359, opened 2018-10-29
File endpoints

mattwiller:


boxcla:

boxsdk/object/file.py
```diff
@@ -79,34 +79,66 @@ def _get_accelerator_upload_url_for_update(self):
...
+            (`int`, `int`)
         :returns:
             File content as bytes.
         :rtype:
             `bytes`
         """
         url = self.get_url('content')
-        box_response = self._session.get(url, expect_json_response=False)
+        params = {'version': file_version.object_id} if file_version is not None else None
+        headers = {'Range': 'bytes={0}-{1}'.format(*byte_range)} if byte_range is not None else None
```
jmoldow: Does Box's byte range support open-ended ranges, like "200-" or "-200"? If so, does the Box SDK want to support it?

I think somewhere on my laptop I have work from when I started working on this feature a few years ago. I was going to create a `ByteRange` object that could support only an upper bound, only a lower bound, or both. And then you would manually construct that object, and pass it in here. The object would have a method that knew how to create the Range header. I could try to dig up that code if you're interested in supporting that in a nice way.

We could also say that, if you pass a 1-tuple, that we assume it means "-{0}". That would match the behavior of Python's `range()` builtin.

boxsdk/object/file.py
```diff
@@ -79,34 +79,66 @@ def _get_accelerator_upload_url_for_update(self):
...
+            :class:`FileVersion` or None
+        :param byte_range:
+            A tuple of inclusive byte offsets to download, e.g. (100, 199) to download the second 100 bytes of a file
+        :type byte_range:
+            (`int`, `int`)
         """
         url = self.get_url('content')
-        box_response = self._session.get(url, expect_json_response=False, stream=True)
+        params = {'version': file_version.object_id} if file_version is not None else None
+        headers = {'Range': 'bytes={0}-{1}'.format(*byte_range)} if byte_range is not None else None
```
jmoldow: We shouldn't to duplicate this logic, we should have a class for it (as I described above) or at least a helper method in this class.

boxsdk/object/file.py
```diff
@@ -433,3 +471,199 @@ def get_tasks(self, fields=None):
...
+        :type etag:
+            `unicode` or None
+        :returns:
+            Whether the operation succeeded.
+        :rtype:
+            `boolean`
+        """
+        url = self.get_url('versions', file_version.object_id)
+        headers = {'If-Match': etag} if etag is not None else None
+        response = self._session.delete(url, expect_json_response=False, headers=headers)
```
jmoldow: This should just be `file_version.delete()`

Do we really need this method?

boxsdk/pagination/marker_based_dict_collection.py
```diff
@@ -0,0 +1,10 @@
+# coding: utf-8
+
+from __future__ import unicode_literals, absolute_import
+
+from .dict_page import DictPage
+from .marker_based_object_collection import MarkerBasedObjectCollection
+
+
+class MarkerBasedDictCollection(MarkerBasedObjectCollection):
```
jmoldow: Docstring

boxsdk/object/file.py
mattwiller: This is the same as in the Email Alias deletion; the URL requires both the file ID and the version ID.

boxsdk/object/file.py
mattwiller: The API only supports an open-ended range like `200-`, not `-200` (which should be expressed as `0-200`)

jmoldow: Oh yeah, duh.

Do we want to support 1-tuple `(200,)` meaning `200-` ?

Or, if we want to force it to always be a 2-tuple, it can be expressed as `(200, None)`.

I guess with your current code, there's a hacky way to get the open-ended range, by passing `(200, '')`. I don't know if we want to officially support/recommend that though.

PR #376, opened 2018-12-07
Stop respecting the 'offset' value in HTTP response.

potrebic:
Paging APIs return total_count, limit, and offset value. The
V2 API docs talk about how to respect the total_count and
limit fields. But do not state that 'offset' should be interpreted.
So remove the ode that was respecting this value as it can lead
to undefined behavior if it starts to change.

Also add test coverage for existing logic in the paging code.

And fix a broken test in test_group where it was passing params
in the wrong order to Group.get_memberships(). A strange
coincidence that these tests were ever passing???


boxcla:

PR #378, opened 2018-12-11
working resume method

carycheng:


boxcla:

boxsdk/util/chunked_uploader.py
```diff
@@ -31,6 +33,8 @@ def __init__(self, upload_session, content_stream, file_size):
         self._file_size = file_size
         self._part_array = []
         self._sha1 = hashlib.sha1()
+        self._part_definitions = defaultdict(list)
```
mattwiller: Why is this a `defaultdict(list)`?  When the key isn't found, this will return an empty list; is that the desired behavior?

boxsdk/util/chunked_uploader.py
```diff
@@ -31,6 +33,8 @@ def __init__(self, upload_session, content_stream, file_size):
...
+        self._is_read = False
```
mattwiller: I don't think you need this anymore; it's not being read anywhere

boxsdk/util/chunked_uploader.py
```diff
@@ -46,17 +50,41 @@ def start(self):
...
+
     def _upload(self):
         """
         Utility function for looping through all parts of of the upload session and uploading them.
         """
         while len(self._part_array) < self._upload_session.total_parts:
             next_part = self._inflight_part or self._get_next_part()
             self._inflight_part = next_part
-            uploaded_part = self._inflight_part.upload()
+            uploaded_part = self._part_definitions[next_part.offset]
```
mattwiller: You could just use `self._part_definitions.get(next_part.offset)` here if you didn't want to make the entire mapping a `defaultdict`

boxsdk/util/chunked_uploader.py
```diff
@@ -72,6 +100,7 @@ def _get_next_part(self):
         offset = len(self._part_array) * self._upload_session.part_size
         while copied_length < self._upload_session.part_size:
             bytes_read = self._content_stream.read(self._upload_session.part_size - copied_length)
+            self._is_read = True
```
mattwiller: You can probably safely remove this, since the value isn't being read anywhere

docs/usage/files.md
```diff
@@ -217,6 +217,28 @@ print('File "{0}" uploaded to Box with file ID {1}'.format(uploaded_file.name, u
...
+Sometimes an upload can be interrupted, in order to resume uploading where you last left off, simply call the
+[`chunked_uploader.resume()`][resume] method. This will return the the [File][file_class] object that was uploaded.
+
+```python
+chunked_uploader = client.file('12345').get_chunked_uploader('/path/to/file')
+uploaded_file = chunked_uploader.resume()
+print('File "{0}" uploaded to Box with file ID {1}'.format(uploaded_file.name, uploaded_file.id))
+```
+
+Alternatively, you can also create a [`UploadSession`][upload_session_class] object by calling
```
mattwiller: It might be good to explain exactly why the user would want to do this (i.e. in order to resume an existing upload session)

docs/usage/files.md
```diff
@@ -217,6 +217,28 @@ print('File "{0}" uploaded to Box with file ID {1}'.format(uploaded_file.name, u
 [get_chunked_uploader]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.upload_session.UploadSession.get_chunked_uploader
 [get_chunked_uploader_for_stream]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.upload_session.UploadSession.get_chunked_uploader_for_stream
 
+#### Resume Upload
+
+Sometimes an upload can be interrupted, in order to resume uploading where you last left off, simply call the
+[`chunked_uploader.resume()`][resume] method. This will return the the [File][file_class] object that was uploaded.
+
+```python
```
mattwiller: We should probably flesh out this example, since it doesn't really make sense as written — why would the user call resume on a new `ChunkedUploader`?  Maybe show a try/except with a call to `start()` that raises an error and a call to `resume()` in the except block?

test/unit/object/test_chunked_upload.py
```diff
@@ -134,3 +137,104 @@ def test_start(test_upload_session, mock_box_session):
...
+            }
+        ]
+    }
+    mock_iterator.__iter__.return_value = [first_part, fourth_part]
+    upload_session_mock_object.get_parts.return_value = mock_iterator
+    chunked_uploader = ChunkedUploader(upload_session_mock_object, stream, file_size)
+    chunked_uploader.resume()
+    calls = [call(offset=2, part_bytes=b'cd', total_size=7),
+             call(offset=4, part_bytes=b'ef', total_size=7), ]
+    upload_session_mock_object.upload_part_bytes.assert_has_calls(calls, any_order=False)
```
mattwiller: This test should also assert that the upload is committed and have some assertions around the return value of `resume()`

test/unit/object/test_chunked_upload.py
```diff
@@ -134,3 +137,104 @@ def test_start(test_upload_session, mock_box_session):
...
+    upload_session_mock_object.get_parts.return_value = mock_iterator
+    chunked_uploader = ChunkedUploader(upload_session_mock_object, stream, file_size)
+    upload_session_mock_object.upload_part_bytes.side_effect = [uploaded_part_one_mock, uploaded_part_two_mock,
+                                                                BoxAPIException(502), uploaded_part_four_mock]
+    try:
+        chunked_uploader.start()
+    except BoxAPIException:
+        chunked_uploader.resume()
+    calls = [call(offset=6, part_bytes=b'g', total_size=7)]
+    upload_session_mock_object.upload_part_bytes.assert_has_calls(calls, any_order=False)
```
mattwiller: This test should also assert that the upload is committed and have some assertions around the return value of `resume()`

boxsdk/util/chunked_uploader.py
```diff
@@ -46,17 +47,41 @@ def start(self):
...
+        content_sha1 = self._sha1.digest()
+        return self._upload_session.commit(content_sha1=content_sha1, parts=self._part_array)
+
     def _upload(self):
         """
         Utility function for looping through all parts of of the upload session and uploading them.
         """
         while len(self._part_array) < self._upload_session.total_parts:
             next_part = self._inflight_part or self._get_next_part()
             self._inflight_part = next_part
```
Jeff-Meadows: Now that you have resume functionality, I don't think `_inflight_part` is needed. Instead, you could get the stream position in the constructor, and reset it to the saved position in resume. Then you can just read the chunks as normal.

carycheng: Hey @Jeff-Meadows can you clarify a little more about what you mean by getting the stream position in the constructor and resetting it to the saved position in resume? 

We were operating under the assumption that the stream may not be seakable all the time

PR #383, opened 2018-12-19
Metadata classification v2

carycheng:


boxcla:

boxsdk/object/item.py
```diff
@@ -11,6 +12,13 @@
 from ..pagination.marker_based_object_collection import MarkerBasedObjectCollection
 
 
+class ClassificationType(TextEnum):
```
mattwiller: Does this include all possible values?  It might be good to look at the metadata template and ensure that the useful enum values of that field are all represented here

boxsdk/object/item.py
```diff
@@ -613,3 +621,94 @@ def get_collaborations(self, limit=None, marker=None, fields=None):
...
+            `unicode`
+        """
+        try:
+            classification = self.metadata('enterprise', 'securityClassification-6VMVochwUWo').get()
+        except BoxAPIException as err:
+            if err.status == 404:
+                return None
+            else:
+                raise
+        return classification['Box__Security__Classification__Key']
```
mattwiller: If the classification has been removed, this will raise a `KeyError`!  You should use something like `return classification.get('Box__Security__Classification__Key', None)` to ensure that a value is always returned

boxsdk/object/item.py
```diff
@@ -613,3 +621,94 @@ def get_collaborations(self, limit=None, marker=None, fields=None):
...
+    def remove_classification(self):
+        """
+        Removes a metadata classification from a :class:`File` or :class:`Folder`.
+
+        :returns:
+            Whether or not the delete was successful.
+        :rtype:
+            `bool`
+        """
+        return self.metadata('enterprise', 'securityClassification-6VMVochwUWo').delete()
```
mattwiller: This will delete the entire template; we had discussed aligning with what the webapp does by deleting only the field containing the classification and leaving the template in place.

Also, how should we handle cases where there was already no classification?

boxsdk/object/item.py
```diff
@@ -613,3 +621,94 @@ def get_collaborations(self, limit=None, marker=None, fields=None):
...
+        :type classification:
+            `unicode`
+        :return:
+            The classification updated on the :class:`File` or :class:`Folder.
+        :rtype:
+            `unicode`
+        """
+        classification_metadata = self.metadata('enterprise', 'securityClassification-6VMVochwUWo')
+        updates = classification_metadata.start_update()
+        updates.update('/Box__Security__Classification__Key', classification)
```
mattwiller: Does this always work?  I think there are two possible cases for updating the metadata instance:

1. The instance already has a value for the classification, so update is correct
2. The instance exists, but does not already have the `Box__Security__Classification__Key` field set — like when the user has used the webapp to remove classification — so update may not work (you may need add instead)

boxsdk/object/item.py
```diff
@@ -613,3 +621,94 @@ def get_collaborations(self, limit=None, marker=None, fields=None):
...
+
+        :return:
+            The classification on the :class:`File` or :class:`Folder.
+        :rtype:
+            `unicode`
+        """
+        try:
+            classification = self.metadata('enterprise', 'securityClassification-6VMVochwUWo').get()
+        except BoxAPIException as err:
+            if err.status == 404:
```
mattwiller: What happens if the user does something like this?

```python
file = client.file('id_that_does_not_exist').get_classification()
```

I think the 404 could apply to either the file itself or the classification — we would probably want the former case to still throw, but the latter to return `None`

boxsdk/object/item.py
```diff
@@ -613,3 +621,94 @@ def get_collaborations(self, limit=None, marker=None, fields=None):
...
+            `unicode`
+        :return:
+            The classification added to the :class:`File` or :class:`Folder.
+        :rtype:
+            `unicode`
+        """
+        classification_metadata = {
+            'Box__Security__Classification__Key': classification,
+        }
+        template_key = 'securityClassification-6VMVochwUWo'
```
mattwiller: It might be worth pulling the template key and field key out into constants so you don't need to keep repeating them everywhere (and so they're easy to update in the future if we need to)

docs/usage/files.md
```diff
@@ -761,3 +762,57 @@ for instance in file_metadata:
 ```
 
 [get_all_metadata]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.item.Item.get_all_metadata
+
+Set a Classification
```
mattwiller: It might be good to note that this functionality is only available to customers with Governance

docs/usage/files.md
```diff
@@ -761,3 +762,57 @@ for instance in file_metadata:
...
+
+Alternatively, if you already know you have a classification and you are simple updating it, you can use the 
+[`update_classification(classification)`][update_classification]. This will ultimately help you save one extra API call.
+
+```python
+classification = client.file(file_id='11111').update_classification(ClassificationType.NONE)
+print('Classification Type is: '.format(classification))
+```
+
+[add_classification]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.item.Item.set_classification
```
mattwiller: I think this reference name is mismatched with the docs above — it says add, but points to set

docs/usage/folders.md
```diff
@@ -301,3 +301,57 @@ for item in items:
             metadata['vendorName'],
         ))
 ```
+
+Set a Classification
```
mattwiller: The comments on the Files docs also apply here

boxsdk/object/item.py
carycheng: @mattwiller in this case do we try update first and if that throws we catch and do add or do we just completely do add instead?

boxsdk/object/item.py
carycheng: @mattwiller in this case are you referring to:

```python
if err.status == 404
	raise
else:
	return None
```

or are you referring to:

1.) Checking if the file.get() is a 404
2.) Throw if file.get() returns 404
3.) Return None if file.get().get_classification() returns 404.

If latter, what is a good way to check for file.get() is 404 outside of making an explicit call to check if file exists first?

PR #392, opened 2019-01-16
Add quotes around JWT install package name to avoid package not found…

jcleblanc:
Related error description and resolution: https://stackoverflow.com/questions/42915098/pip-install-boxsdk-with-jwt-oath-not-functioning-as-expected


boxcla:
Verified that @jcleblanc has signed the CLA. Thanks for the pull request!

PR #394, opened 2019-01-24
Response Object unfurling no longer throws when encountering object.lock

### LINKED ISSUE #390: Event object to python dict using`response_object` method. Throws TypeError: can't pickle _thread.lock objects

### Description of the Issue

Since the event object no longer returns a python dict and rather an Event object. I am trying turn this event object to a python dict.

The method `response_object()` given to an [event object](https://github.com/box/box-python-sdk/blob/master/boxsdk/object/base_api_json_object.py) does not convert the `event object` to a Python `dict` as stated in the 2.0 [release notes](https://github.com/box/box-python-sdk/releases) (under the search query: `response_object`) and instead throws a thread_lock error.

Looking at this it seems to be an issue on the Python package `copy`'s end or perhaps I may be using the `response_object()` method incorrectly.

### Versions Used

Box Python SDK: 2.0
Python: 3.6.1

### Steps to Reproduce

```python
def events_to_json(spoof_as_user):
    """
    Turn event object to python dict for json.dumps usage

    :param spoof_as_user:
        Client acting as a user using "as_user" method
    """

    # Retrieve all user events as iteratable
    all_events = spoof_as_user.events().get_events()

    # Grab first event within the events iteratable
    event_obj = all_events['entries'][0]

    # Format event object to a python dictionary
    print("Turn event object JSON readable using response_method(): {0} ".format(event_obj.response_object()))

```

```
  File "example.py", line 576, in <module>
    main()
  File "example.py", line 565, in main
    events_to_json(user_01_as_user)
  File "example.py", line 505, in events_to_json
    print("Turn event object JSON readable using response_method(): {0} ".format(event_obj.response_object()))
  File "/Users/example/myENV/lib/python3.6/site-packages/boxsdk/object/base_api_json_object.py", line 156, in response_object
    return copy.deepcopy(self._response_object)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Users/example/myENV/lib/python3.6/copy.py", line 169, in deepcopy
    rv = reductor(4)
TypeError: can't pickle _thread.lock objects
```

carycheng:
Fixes #390

### DIFF
--- a/boxsdk/object/base_api_json_object.py
+++ b/boxsdk/object/base_api_json_object.py
@@ -153,20 +153,4 @@ class BaseAPIJSONObject(object):
         :rtype:
             `dict`
         """
-        def untranslate(value):
-            if isinstance(value, BaseAPIJSONObject):
-                copied_object = value.response_object
-                return copied_object.response_object
-            if isinstance(value, dict):
-                return untranslate(value)
-            if isinstance(value, list):
-                copied_object = copy.copy(value.pop(0))
-                values = copied_object.values()
-                return values.response_object
-            else:
-                return value
-
-        copied_object = copy.copy(self._response_object)
-        for key, value in copied_object.items():
-            copied_object[key] = untranslate(value)
-        return copied_object
+        return Translator._default_translator.untranslate(self)
--- a/boxsdk/object/base_api_json_object.py
+++ b/boxsdk/object/base_api_json_object.py
@@ -158,14 +158,11 @@ class BaseAPIJSONObject(object):
             A dictionary containing the untranslated object.
         """
         if isinstance(value, BaseAPIJSONObject):
-            new_value = cls._untranslate(value._response_object)  # pylint:disable=protected-access
-            return new_value
+            return cls._untranslate(value._response_object)
         if isinstance(value, dict):
-            new_dict = {k: cls._untranslate(v) for (k, v) in six.iteritems(value)}
-            return new_dict
+            return {k: cls._untranslate(v) for (k, v) in six.iteritems(value)}
         if isinstance(value, list):
-            new_list = [cls._untranslate(entry) for entry in value]
-            return new_list
+            return [cls._untranslate(entry) for entry in value]
 
         return copy.copy(value)
 
@@ -177,4 +174,4 @@ class BaseAPIJSONObject(object):
         :rtype:
             `dict`
         """
-        return self._untranslate(self)  # pylint:disable=protected-access
+        return self._untranslate(self)
--- a/boxsdk/object/base_api_json_object.py
+++ b/boxsdk/object/base_api_json_object.py
@@ -153,8 +153,20 @@ class BaseAPIJSONObject(object):
         :rtype:
             `dict`
         """
+        def untranslate(value):
+            if isinstance(value, BaseAPIJSONObject):
+                copied_object = value.response_object
+                return copied_object.response_object
+            if isinstance(value, dict):
+                return untranslate(value)
+            if isinstance(value, list):
+                copied_object = copy.copy(value.pop(0))
+                values = copied_object.values()
+                return values.response_object
+            else:
+                return value
+
         copied_object = copy.copy(self._response_object)
         for key, value in copied_object.items():
-            if isinstance(value, BaseAPIJSONObject):
-                copied_object[key] = value.response_object
+            copied_object[key] = untranslate(value)
         return copied_object
--- a/boxsdk/object/base_api_json_object.py
+++ b/boxsdk/object/base_api_json_object.py
@@ -158,7 +158,7 @@ class BaseAPIJSONObject(object):
             A dictionary containing the untranslated object.
         """
         if isinstance(value, BaseAPIJSONObject):
-            return cls._untranslate(value._response_object)  #pylint:disable=protected-access
+            return cls._untranslate(value._response_object)  # pylint:disable=protected-access
         if isinstance(value, dict):
             return {k: cls._untranslate(v) for (k, v) in six.iteritems(value)}
         if isinstance(value, list):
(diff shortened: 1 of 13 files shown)


boxcla:

boxsdk/util/translator.py
```diff
@@ -191,5 +193,28 @@ def translate(self, session, response_object):
 
         return translated_obj
 
+    def untranslate(self, value):
```
jmoldow: This doesn't need to be a method on the `Translator` class. It doesn't actually use any of the other methods or data of the `Translator`. Additionally, we don't have a `Translator` instance available (besides the `_default_translator`) in the place where we need to use this method.

This can be either a `@classmethod` for `Translator`, or it can be a private `@classmethod` on `BaseAPIJSONObject`. I somewhat prefer the latter, because then the logic for `response_object` exists in a single place, and because it doesn't involve exposing a new public method that we'll need to maintain.

boxsdk/object/base_api_json_object.py
```diff
@@ -153,4 +152,4 @@ def response_object(self):
         :rtype:
             `dict`
         """
-        return copy.deepcopy(self._response_object)
+        return Translator._default_translator.untranslate(self)  # pylint:disable=protected-access
```
jmoldow: We shouldn't ever use the default translator for performing translations.

boxsdk/util/translator.py
```diff
@@ -191,5 +193,28 @@ def translate(self, session, response_object):
...
+
+        :param value:
+            The object to untranslate.
+        :rtype value:
+            `dict`
+        :return:
+            A dictionary containing the untranslated object.
+        """
+        if hasattr(value, '_response_object') and isinstance(value._response_object, dict):  # pylint:disable=protected-access
+            new_value = self.untranslate(value._response_object)  # pylint:disable=protected-access
```
jmoldow: You can instead set `value = ` and get rid of the return statement. That way it'll fall-through to the `if isinstance(value, dict)`, rather than needing to do a recursion.

boxsdk/util/translator.py
```diff
@@ -191,5 +193,28 @@ def translate(self, session, response_object):
...
+        Untranslates a given object into a dictionary.
+
+        :param value:
+            The object to untranslate.
+        :rtype value:
+            `dict`
+        :return:
+            A dictionary containing the untranslated object.
+        """
+        if hasattr(value, '_response_object') and isinstance(value._response_object, dict):  # pylint:disable=protected-access
```
jmoldow: Another reason to put this on `BaseAPIJSONObject`: because you're accessing one of its private fields.

Rather than do `hasattr(value, '_response_object')`, it would be stronger to do `isinstance(value, BaseAPIJSONObject)`

boxsdk/object/base_api_json_object.py
```diff
@@ -145,6 +145,30 @@ def untranslated_fields(cls):
...
+
+        :param value:
+            The object to untranslate.
+        :rtype value:
+            `dict`
+        :return:
+            A dictionary containing the untranslated object.
+        """
+        if isinstance(value, BaseAPIJSONObject):
+            new_value = cls._untranslate(value._response_object)  # pylint:disable=protected-access
```
mattwiller: You shouldn't need to disable pylint here anymore

boxsdk/object/base_api_json_object.py
```diff
@@ -145,6 +145,30 @@ def untranslated_fields(cls):
...
+            return new_value
```
mattwiller: It might be nice to one-line all of these instead of declaring a new variable and then instantly returning it

boxsdk/object/base_api_json_object.py
```diff
@@ -153,4 +177,4 @@ def response_object(self):
         :rtype:
             `dict`
         """
-        return copy.deepcopy(self._response_object)
+        return self._untranslate(self)  # pylint:disable=protected-access
```
mattwiller: I think you shouldn't need to disable pylint here, since we're in this class

PR #402, opened 2019-02-11
Change api request retry strategy to use exponential backoff

carycheng:


boxcla:

boxsdk/session/session.py
```diff
@@ -242,8 +249,11 @@ def with_default_network_request_kwargs(self, extra_network_parameters):
         kwargs['default_network_request_kwargs'].update(extra_network_parameters)
         return self.__class__(**kwargs)
 
+    # pylint: disable=unused-argument
```
mattwiller: Could you add a comment here to explain why this is being done?  We wouldn't want someone else to see this and think the unused argument can just be removed

boxsdk/session/session.py
```diff
@@ -242,8 +249,11 @@ def with_default_network_request_kwargs(self, extra_network_parameters):
...
     def _get_retry_after_time(self, attempt_number, retry_after_header):
         """
+        We are using the attempt number that failed to calculate the retry time for the next retry attempt.
```
mattwiller: This comment should probably be integrated with the broader description of this method, maybe something like this:

`Get the amount of time to wait before retrying the API request, using the attempt number that failed to calculate the retry time for the next retry attempt.`

test/functional/test_rate_limits.py
```diff
@@ -2,9 +2,12 @@
...
 from __future__ import unicode_literals
 
+from mock import patch
+
 
 def test_too_many_requests_causes_retry(box_client, mock_box, monkeypatch):
     monkeypatch.setattr(mock_box, 'RATE_LIMIT_THRESHOLD', 1)
-    box_client.folder('0').get()
-    box_client.folder('0').get()
+    with patch('random.uniform', return_value=1):
```
mattwiller: You should also leave a comment here on why it's important to have the randomness stubbed out to the highest possible value (to avoid the race condition)

test/functional/test_recovery.py
```diff
@@ -66,7 +66,7 @@ def test_client_stops_retrying_after_10_server_errors(box_client, mock_box, erro
     with pytest.raises(BoxAPIException) as exc_info:
         box_client.folder('0').get()
         assert exc_info.value.status == error_code
-    assert len(mock_box.requests) == 14  # 3 auth requests, 1 try, and 10 retries
+    assert len(mock_box.requests) == 9  # 3 auth requests, 1 try, and 10 retries
```
mattwiller: The comment here should be adjusted to match the assertion

test/unit/session/test_session.py
```diff
@@ -145,11 +145,12 @@ def test_box_session_retries_response_after_retry_after(
...
     mock_network_layer.retry_after.side_effect = lambda delay, request, *args, **kwargs: request(*args, **kwargs)
 
-    box_response = test_method(box_session, url=test_url)
+    with patch('random.uniform', return_value=0.68):
+        box_response = test_method(box_session, url=test_url)
     assert box_response.status_code == 200
     assert len(mock_network_layer.retry_after.call_args_list) == 1
     assert isinstance(mock_network_layer.retry_after.call_args[0][0], Number)
-    assert mock_network_layer.retry_after.call_args[0][0] == float(retry_after_response.headers['Retry-After'])
+    assert mock_network_layer.retry_after.call_args[0][0] == 1.1800000000000002
```
mattwiller: This assertion should be truncated like the others

test/unit/session/test_session.py
```diff
@@ -172,16 +173,17 @@ def test_box_session_retries_request_after_server_error(
...
     assert box_response.json() == generic_successful_response.json()
     assert box_response.ok == generic_successful_response.ok
     assert box_response.content == generic_successful_response.content
     assert len(mock_network_layer.retry_after.call_args_list) == 2
     assert isinstance(mock_network_layer.retry_after.call_args_list[0][0][0], Number)
     assert isinstance(mock_network_layer.retry_after.call_args_list[1][0][0], Number)
-    assert mock_network_layer.retry_after.call_args_list[0][0][0] == 1
-    assert mock_network_layer.retry_after.call_args_list[1][0][0] == 2
+    assert mock_network_layer.retry_after.call_args_list[0][0][0] == 1.1800000000000002
+    assert mock_network_layer.retry_after.call_args_list[1][0][0] == 2.3600000000000003
```
mattwiller: These assertions should be truncated like the others

PR #407, opened 2019-03-05
Basic proxy support in Session

carycheng:


boxcla:

boxsdk/session/session.py
```diff
@@ -35,6 +35,7 @@ def __init__(
             default_network_request_kwargs=None,
             api_config=None,
             client_config=None,
+            network_config=None,
```
mattwiller: This parameter needs to be documented in the docstring below

boxsdk/session/session.py
```diff
@@ -184,6 +186,14 @@ def client_config(self):
         """
         return self._client_config
 
+    @property
+    def network_config(self):
+        """
+
+        :rtype:     :class:`Network`
```
mattwiller: It's a bit weird that this config class has a name conflict with the `Network` interface that `DefaultNetwork` implements — we may want to disambiguate them

boxsdk/session/session.py
```diff
@@ -425,6 +435,28 @@ def _get_retry_request_callable(self, network_response, attempt_number, request)
...
+            A prepared proxy dict to send along with the request.
+        :rtype:
+            `dict`
+        """
+        proxy = {}
+        if self._network_config.PROXY_AUTH and self._network_config.PROXY_URL:
+            host = self._network_config.PROXY_URL
+            address = host.split('//')[1]
+            proxy['http'] = \
+                'http://' + self._network_config.PROXY_AUTH.get('user', None) + \
```
mattwiller: It might be clearer to use a format string here to show exactly what you're trying to get the resulting string to look like

test/unit/session/test_session.py
```diff
@@ -293,3 +293,35 @@ def test_get_retry_after_time(box_session, attempt_number, retry_after_header, e
         retry_time = box_session._get_retry_after_time(attempt_number, retry_after_header)  # pylint: disable=protected-access
     retry_time = round(retry_time, 4)
     assert retry_time == expected_result
+
+
+@pytest.mark.parametrize(
+    'test_proxy_url,test_proxy_auth, expected_proxy_dict',
+    [
+        ('http://example-proxy.com', {'user': 'test_user', 'password': 'test_password', },
```
mattwiller: What happens if the user only provides one of the expected keys in the `PROXY_AUTH` dict?  Or if they put something unrelated in there (e.g. `{ 'foo': 'bar' }`)?

Issue #408, opened 2019-03-06
boxsdk 2.2.1 is incompatible with requests-toolbelt 0.9.1 


breakaway83:
- [X] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
boxsdk 2.2.1 has requirement requests-toolbelt<0.9.0,>=0.4.0, but you'll have requests-toolbelt 0.9.1 which is incompatible.

### Versions Used
Box Python SDK: 2.2.1  
Python: 2.7.15

### Steps to Reproduce
requirements.txt:
boxsdk>=1.5.3
requests-toolbelt>=0.4.0

pip install -r requirements.txt

### Error Message, Including Traceback
boxsdk 2.2.1 has requirement requests-toolbelt<0.9.0,>=0.4.0, but you'll have requests-toolbelt 0.9.1 which is incompatible.

[sdk-docs]: ./docs/usage
[api-docs]: https://developer.box.com/docs

carycheng:
Hey @breakaway83,

Just to clarify, do you mean the SDK is pulling in requests-toolbelt 0.9.1 when you install or are you using requests-toolbelt 0.9.1 and need support for it in the sdk?

Thanks!

breakaway83:
When I build boxsdk together with requests-toolbelt, it shows the warning.
I guess it is the latter:
"using requests-toolbelt 0.9.1 and need support for it in the sdk"

Weimin

carycheng:
Hey @breakaway83,

We've filed a ticket for this in our backlog and will keep you updated once we've completed this!

breakaway83:
Thanks.

Weimin

carycheng:
Hey @breakaway83,

Thank you for your patience! We just release v2.2.2 of the boxsdk with updated restrictions to allow requests-toolbelt v0.9.1.

Please check it out and let us know if you run into any issues, thanks!

breakaway83:
Thanks. Looks like it is working now.

Weimin

carycheng:
@breakaway83,

Glad to hear Weimin! I'm going to close this ticket now.

If you come across any other issues please do not hesitate to reach out.

Thanks!

PR #410, opened 2019-03-10
Upgrading request-toolbelt to accept 0.9.1

carycheng:
It looks like in the changelog for request-toolbelt 0.9.1 there was a bug fix for importing pyOpenSSL.

This previously was already broken so it should not affect the SDK.

#408

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,7 +58,7 @@ def main():
     install_requires = [
         'attrs>=17.3.0',
         'requests>=2.4.3',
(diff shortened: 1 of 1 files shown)


boxcla:

mattwiller:
Now that we're back into a working state with the current versions of this library, it might be good to just make the version range open-ended again and take any version of `requests-toolbelt` >= 0.4.0 so we don't have people asking us to update again later.  For maximum safety, we could always do `>=0.4.0,<1.0.0` to get any new minor versions

PR #421, opened 2019-04-23
Added support for setting metadata

carycheng:


boxcla:

boxsdk/object/metadata.py
```diff
@@ -206,6 +207,33 @@ def create(self, metadata):
...
+            A dictionary containing the key/value pairs for this metadata object.
+        :rtype:
+            :class:`Metadata`
+        """
+        metadata_value = None
+        try:
+            metadata_value = self.create(metadata)
+        except BoxAPIException as err:
+            if err.status == 409:
+                metadata_instance = Metadata(self._session, self, self._scope, self._template)
```
mattwiller: Isn't this object (`self`) already a `Metadata` instance?  Could you just call `self.start_update()` instead?

docs/usage/files.md
```diff
@@ -689,6 +696,25 @@ print('Applied metadata in instance ID {0}'.format(applied_metadata['$id']))
 [metadata]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.item.Item.metadata
 [metadata_create]: https://box-python-sdk.readthedocs.io/en/latest/boxsdk.object.html#boxsdk.object.metadata.Metadata.create
 
+Set Metadata
```
mattwiller: As I mentioned in the Java SDK pull request, I think we should condense the existing documentation for "create" and "update" metadata into this section as additional information.  This new method should be the default for most developers, and it would be helpful if the table of contents above didn't have three sections for them to pick from when trying to figure out how to set metadata!

docs/usage/files.md
```diff
@@ -36,10 +39,14 @@ file's contents, upload new versions, and perform other common file operations
 - [Get File Representations](#get-file-representations)
 - [Get Thumbnail](#get-thumbnail)
 - [Add Metadata](#add-metadata)
+- [Set Metadata](#set-metadata)
 - [Get Metadata](#get-metadata)
 - [Update Metadata Values](#update-metadata-values)
 - [Remove Metadata](#remove-metadata)
 - [Get All Metadata](#get-all-metadata)
+- [Set a Classification](#set-a-classification)
```
mattwiller: Speaking of these methods, you might be able to update the implementation of "Set Classification" to use the new set metadata method now

PR #434, opened 2019-06-19
Limit line length for get_folders_id

cbetta:
Reduced the line length for the `get_folders_id` sample in order to make it render nicely on box.dev

### DIFF
--- a/docs/usage/folders.md
+++ b/docs/usage/folders.md
@@ -43,7 +43,8 @@ ones not returned by default.
 <!-- sample get_folders_id -->
 ```python
 folder = client.folder(folder_id='22222').get()
-print('Folder "{0}" has {1} items in it'.format(folder.name, folder.item_collection.total_count))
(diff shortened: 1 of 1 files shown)


boxcla:

docs/usage/folders.md
```diff
@@ -43,7 +43,8 @@ ones not returned by default.
 <!-- sample get_folders_id -->
 ```python
 folder = client.folder(folder_id='22222').get()
-print('Folder "{0}" has {1} items in it'.format(folder.name, folder.item_collection.total_count))
+print('Folder "{0}" has {1} items in it'
```
Jeff-Meadows: our general Python coding style would have this line as 

    print('Folder "{0}" has {1} items in it'.format(
        folder.name,
        folder.item_collection.total_count,
    ))

cbetta: That is actually even better. I'll send an update.

Issue #439, opened 2019-07-31
Pull Events by Created Date and/or Event Type


capk1rk:
### Description of the Issue
The current get_events() function pulls all events, using stream position to page through the data _n_ number at a time where _n_ is the set limit. 

### Versions
Box Python SDK: boxsdk 2.5.0
Python: Python 3.6.8

### Solution
Currently dynamically overriding this method at runtime to add `created_after` and `event_type`. Thought it would be useful for others, so I decided to add a new function with this code. Unfortunately, paging with `stream_position` does not work with either of those parameters, so that will be left out. 

### Related
#257

carycheng:
@capk1rk 

Thank you for the pr! We'll take a look at this and update you

Much appreciated!

carycheng:
Hello @capk1rk,

Thank you for the pr!

We will review this with the team.

Much appreciated!

carycheng:
Hey @capk1rk,

Thank you very much for your patience. Your enhancement has been merged at #441 
We really appreciate the work you did to better our developer tool. If you run across any other areas in the future you see fit to improve, please do not hesitate to reach out to us.

PR #257, opened 2018-02-02
Added a new method 'get_events_by_date'

Uppu-Srikanth:
This helps in fetching events between two specified timestamps/dates with/without stream position mentioned.
If stream position is not mentioned , it returns the events from the start of the 'created_after' timestamp , else it returns all the events between 'created_after' and 'created_before' timestamp

PR #440, opened 2019-08-02
Add get event by details

capk1rk:
Added a function with the ability to pull events with additional parameters. Almost identical to the original get_events() function. 
- `created_after` Only return events that happen after this date
- `created_before` Only return events that happen before this date
- `event_type` Type of events to limit response too

All of the parameters are optional. All of these parameters are incompatible with stream positions. Max limit of 500. The Box API documentation says that `created_after` and `created_before` parameters only work with admin logs. From my testing, it works with all stream types.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -9,3 +9,4 @@ used to interact with the Box API. This is a list of contributors.
 - `@nsundareswaran <https://github.com/nsundareswaran>`_
 - `@kelseymorris95 <https://github.com/kelseymorris95>`_
 - `@sp4x <https://github.com/sp4x>`_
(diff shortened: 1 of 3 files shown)


boxcla:
Hi @capk1rk, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

capk1rk:
CLA Signed. 
Linked to issue #439 and pull request #257

boxcla:
Verified that @capk1rk has just signed the CLA. Thanks, and we look forward to your contribution.

PR #450, opened 2019-10-08
Ensure that token exchange works before an API call is made

mattwiller:
Previously, the client.downscope_token() method assumed that an
access token was available, but for JWT auth applications a token
is not generated until the first API call is made; this causes
token exchange to fail.  The downscope method now attempts to
generate an access token if one is not present, before performing
the token exchange.


boxcla:

PR #451, opened 2019-10-09
Update authentication.md

cbetta:
Added sample IDs to all auth samples


boxcla:

PR #453, opened 2019-10-14
Add missing @api_call decorator (fixes #452)

### LINKED ISSUE #452: File.copy() missing api_call decorator
I noticed that `File.copy()`, which overrides `Item.copy()` to add an optional version check, does not have an `@api_call` decorator, which prevents the use of `extra_network_parameters`. Is this omission intentional?

See here:
https://github.com/box/box-python-sdk/blob/ea0ad05e6208fda262199658a2b6290eb9b5a747/boxsdk/object/file.py#L678-L681

Compare with:
https://github.com/box/box-python-sdk/blob/1cec692f01631b158d0de9c97477262e302e1037/boxsdk/object/item.py#L147-L151

nidoran:


boxcla:
Hi @nidoran, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

PR #456, opened 2019-10-19
Update files.md

cbetta:
Add better tags for download samples. Had to split some samples.


boxcla:

PR #458, opened 2019-10-21
Update collections.md

cbetta:
Add variant code sample tags for collections


boxcla:

PR #459, opened 2019-10-22
Update comments.md

cbetta:
Add sample tags for variations of comment usage

### DIFF
--- a/docs/usage/comments.md
+++ b/docs/usage/comments.md
@@ -54,14 +54,21 @@ Add a Comment to a File
 -----------------------
 
 To leave a comment on a file, call [`file.add_comment(message)`][add_comment] with the message to leave in the comment.
-You can at-mention other users by adding special tags within the message, in the format `@[USER_ID:USER_NAME]`.  For
-example, to at-mention John Doe, whose user ID is `"33333"`: `@[33333:John Doe]`.  This method returns a
-[`Comment`][comment_class] object representing the newly-created comment.
 
 <!-- sample post_comments -->
 ```python
+comment = client.file(file_id='11111').add_comment('When should I have this done by?')
+```
+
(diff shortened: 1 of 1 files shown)


boxcla:

docs/usage/comments.md
```diff
@@ -54,29 +54,44 @@ Add a Comment to a File
...
-example, to at-mention John Doe, whose user ID is `"33333"`: `@[33333:John Doe]`.  This method returns a
-[`Comment`][comment_class] object representing the newly-created comment.
 
 <!-- sample post_comments -->
 ```python
+comment = client.file(file_id='11111').add_comment('When should I have this done by?')
+```
+
+You can at-mention other users by adding special tags within the message, in the format `@[USER_ID:USER_NAME]`.  For
+example, to at-mention John Doe, whose user ID is `"33333"`: `@[33333:John Doe]`.
```
PJSimon: Here the User Name is first name + last name.
But in https://github.com/box/box-java-sdk/pull/749/files, the example is:
`comment.reply("@[1234:user@box.com] I agree with this!");`
Are both correct?

PJSimon:
> I have a question. If I'm wrong, go ahead and merge (I approved). If I'm right, lemme know if you need another approval after making a change.

@cbetta

docs/usage/comments.md
cbetta: @PJSimon so my understanding (and I just confirmed this) is that the first part before the colon is the tagged user ID, where the second part after the colon is the text to use in the link to that user. This means that that text can be the user's name, email, pseudonym, nickname, or anything really. We seem to default to "NAME" in the reference docs as a way to say "any name", but maybe we should be more explicit that that text can be anything. Thoughts?

PR #464, opened 2019-10-31
Fixed `get_admin_events` Optional `event_types` Parameter

### LINKED ISSUE #463: get_admin_events Function Fails When Optional event_types Parameter is Omitted

### Description of the Issue
When using the [get_admin_events](https://github.com/box/box-python-sdk/blob/master/boxsdk/object/events.py#99) function, if you do not provide the `event_types` parameter (listed as optional) then it fails trying to pass `None` to a `join()` call [here](https://github.com/box/box-python-sdk/blob/master/boxsdk/object/events.py#L132).

### Versions Used
Box Python SDK: 2.6.1
Python: 3.7.3

### Steps to Reproduce
1. Create an authenticated client object (we're using JWT)
2. Create an `events` object using `client.events()`
3. Attempt to pull admin events using a command similar to `events.get_admin_events(limit=10, created_after='2019-10-31T08:00:00-0000')`. Both of those arguments are optional, but they were my test values, so they're probably worth using to reproduce.

```
>>> events.get_admin_events(limit=10, created_after='2019-10-31T08:00:00-0000')
  File "<console>", line 1, in <module>
  File "/path/to/project/venv/lib/python3.7/site-packages/boxsdk/util/api_call_decorator.py", line 71, in call
    return method(*args, **kwargs)
  File "/path/to/project/venv/lib/python3.7/site-packages/boxsdk/object/events.py", line 132, in get_admin_events
    'event_type': ','.join(event_types),
TypeError: can only join an iterable
```

[sdk-docs]: ./docs/usage
[api-docs]: https://developer.box.com/docs

aculler:
Per #463, I updated the `get_admin_events` function to properly handle the case where the `event_types` optional parameter is omitted.

Fixes #463.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -10,3 +10,4 @@ used to interact with the Box API. This is a list of contributors.
 - `@kelseymorris95 <https://github.com/kelseymorris95>`_
 - `@sp4x <https://github.com/sp4x>`_
 - `@capk1rk <https://github.com/capk1rk>`_
+- `@aculler <https://github.com/aculler>`_
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ---------------
 
+Upcoming
+++++++++
+- Fixed bug in get_admin_events function which caused errors when the optional event_types parameter was omitted.
+
 2.6.1 (2019-10-24)
 ++++++++++++++++++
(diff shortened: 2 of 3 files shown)


boxcla:
Verified that @aculler has signed the CLA. Thanks for the pull request!

PR #468, opened 2019-12-18
Add marker based pagination for listing users

sujaygarlanka:
### Issue Link :link:
- https://jira.inside-box.net/browse/SDK-1215

### Goals :soccer:
- Add marker based pagination for listing users

### Implementation Details :construction:
- Added `use_marker` and `marker` parameters to users method in `client.py`
- Return marker paginator if `use_marker` parameter is true

### Testing Details :mag:
- Added unit test

### DIFF
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -493,7 +493,7 @@ class Client(Cloneable):
         )
 
     @api_call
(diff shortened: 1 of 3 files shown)


boxcla:

PR #473, opened 2020-01-07
Support additional file attributes on upload (#472)

### LINKED ISSUE #472: Add support for arbitrary attributes for file uploads and updates

### Description of the Issue
The upload methods in the SDK do not support modifying arbitrary file attributes. For example, the [API reference](https://developer.box.com/en/reference/post-files-content/) indicates that it should be possible to specify a `content_created_at` datetime when uploading a new file, but the SDK does not currently make this possible.

The upload methods are...
- `File.update_contents`
- `File.update_contents_with_stream`
- `Folder.upload`
- `Folder.upload_stream`

### Versions Used
Box Python SDK: 2.6.1
Python: 2.7 and 3.7

### Steps to Reproduce
N/A

N/A

[sdk-docs]: ./docs/usage
[api-docs]: https://developer.box.com/docs

ben-reilly:
Added parameters for all attributes explicitly mentioned in the API reference for uploads*, plus added an `additional_attributes` dict that can be optionally filled with key-value pairs with any other attributes the user wants to modify.

This fixes #472.

*See [here for uploading new files](https://developer.box.com/en/reference/post-files-content/) and [here for uploading new file versions](https://developer.box.com/en/reference/post-files-id-content/).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -11,3 +11,4 @@ used to interact with the Box API. This is a list of contributors.
 - `@sp4x <https://github.com/sp4x>`_
 - `@capk1rk <https://github.com/capk1rk>`_
 - `@aculler <https://github.com/aculler>`_
+- `@ben-reilly <https://github.com/ben-reilly>`_
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,7 @@ Release History
 Upcoming
 ++++++++
 - Fixed bug in get_admin_events function which caused errors when the optional event_types parameter was omitted.
+- Added support for more attribute parameters when uploading new files and new versions of existing files.
 
 2.6.1 (2019-10-24)
 ++++++++++++++++++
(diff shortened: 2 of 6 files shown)


boxcla:

ben-reilly:
I'm seeing some errors in CI, but I don't think I caused them.

1. `test_file::test_lock` fails sometimes with Python 2 because dictionaries aren't returned in a predictable order in Py2, resulting in variations in the JSON representation of the dict. Not sure if you've seen this before or how you'd like to resolve it. One way would be to use the `sort_keys=True` param in `json.dumps()`.
1. The Py3.4 build fails because of some incompatibility with a package:
    `ERROR: PyYAML requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.8`
1. The Py3.5 build fails because it fails to import something in the code coverage reporter:
 `    from coverage.report import Reporter`
 `ImportError: cannot import name 'Reporter'`

Any input on how you'd like to resolve/ignore these?

PJSimon:
Hey, @ben-reilly!  Thanks for this contribution!  We will do the review and take a look at the Travis errors very soon!

Issue #475, opened 2020-01-10
Combine preflight check and accelerator URL lookup for uploads


ben-reilly:
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
When uploading a file, a developer can request...
- to perform a preflight check, and/or
- to use an accelerator URL.

If the developer opts for both, we make two OPTIONS requests to the same endpoint. We can combine the two requests to serve a double purpose in this case.

### Versions Used
Box Python SDK: 2.6.1
Python: 2.7 and 3.7

### Steps to Reproduce
N/A

### Error Message, Including Traceback
N/A

[sdk-docs]: ./docs/usage
[api-docs]: https://developer.box.com/docs

ben-reilly:
#476 was merged, fixing this, so I'll close.

PR #476, opened 2020-01-10
Combine preflight and accelerator calls for uploads (#475)

ben-reilly:
When uploading a file, we can combine the preflight check and the request for an accelerator URL into a single OPTIONS request, since both actions target the same endpoint.


boxcla:

PR #478, opened 2020-01-14
Pin coverage package to fix backwards-incompatibility

ben-reilly:
Will create an issue and add explanation once I can get this to pass. :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,6 +100,7 @@ def main():
         'pytest-cov',
         'pytest-xdist<1.28.0',
         'python-coveralls',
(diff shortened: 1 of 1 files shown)


boxcla:

ben-reilly:
For posterity: the `coverage` package version 5.0 introduced changes that were apparently not backwards compatible, which broke `python-coveralls`. This commit pins the `coverage` version to `<5.0` so that we avoid this incompatibility until it's fixed.

See below for additional details and context:
* https://github.com/z4r/python-coveralls/issues/73
* https://github.com/creare-com/podpac/issues/347

PR #479, opened 2020-01-15
Fix JWT retry logic

sujaygarlanka:


boxcla:

boxsdk/auth/jwt_auth.py
```diff
@@ -239,14 +241,32 @@ def _auth_with_jwt(self, sub, sub_type):
...
+            except BoxOAuthException as ex:
+                jwt_time = None
+                network_response = ex.network_response
+                code = network_response.status_code  # pylint: disable=maybe-no-member
+                box_datetime = self._get_date_header(network_response)
+
+                if attempt_number >= API.MAX_RETRY_ATTEMPTS:
+                    raise ex
+
+                if (code == 429 or code >= 500):
```
PJSimon: Lines 259 and 260 are duplicated (262 and 263).  Can you rework this logic so they can appear once?

test/unit/session/test_session.py
```diff
@@ -150,7 +150,7 @@ def test_box_session_retries_response_after_retry_after(
     assert box_response.status_code == 200
     assert len(mock_network_layer.retry_after.call_args_list) == 1
     assert isinstance(mock_network_layer.retry_after.call_args[0][0], Number)
-    assert round(mock_network_layer.retry_after.call_args[0][0], 4) == 1.18
+    assert round(mock_network_layer.retry_after.call_args[0][0], 4) == 1
```
PJSimon: Can you explain this change, please?  :^)

boxsdk/session/session.py
```diff
@@ -285,11 +286,14 @@ def _get_retry_after_time(self, attempt_number, retry_after_header):
         :return:                        Number of seconds to wait before retrying.
         :rtype:                         `Number`
         """
-        min_randomization = 1 - self._retry_randomization_factor
-        max_randomization = 1 + self._retry_randomization_factor
-        randomization = (random.uniform(0, 1) * (max_randomization - min_randomization)) + min_randomization
-        exponential = math.pow(2, attempt_number)
-        return exponential * self._retry_base_interval * randomization
+        try:
```
PJSimon: Is there a way to check for the presence of `retry_after_header` before doing the `try` statement?  I don't know about Python, but there is usually an overhead associated with relying on the exception to be thrown instead of a condition statement.  And I don't know how often we expect `retry_after_header` to be missing ... is it usually present?

boxsdk/session/session.py
```diff
@@ -285,11 +286,14 @@ def _get_retry_after_time(self, attempt_number, retry_after_header):
...
+            return int(retry_after_header)
```
PJSimon: Do we want to just return this value, as-is?  I would think that we'd want to return the greater of:
- `return exponential * self._retry_base_interval * randomization`
- `int(retry_after_header)`

Also, is `retry_after_header` in units of seconds or is it a datetime or... ?

boxsdk/auth/jwt_auth.py
sujaygarlanka: One option is to put the logic in a method. The other is to have a single if statement that checks everything

boxsdk/session/session.py
sujaygarlanka: We can add an if statement that checks whether the header is there.

boxsdk/auth/jwt_auth.py
PJSimon: Sure, either works!

boxsdk/session/session.py
PJSimon: I'm new to Python, I could be wrong:
https://stackoverflow.com/questions/1835756/using-try-vs-if-in-python

test/unit/session/test_session.py
sujaygarlanka: Before, the retry header was not being returned if it was passed. Once I fixed this bug I had to change the tests to work with this new behavior.

boxsdk/session/session.py
sujaygarlanka: Added an if check. Doesn't make sense to cast if we don't need to.

boxsdk/session/session.py
PJSimon: Could this result in the time between retries not following the exponential backoff in the SDK?  

I'm not clear on how the server determines this value.  If the server provides a value, then it makes sense not to retry BEFORE that time.  But if there are multiple requests being rate-limited, we might want the variability that the SDK's backoff logic provides in order to spread out requests to avoid a Thundering Herd situation.

sujaygarlanka: I assume the server must calculate some variable time that request should wait for after a 429 since it is should be giving a time after which your service should not get another 429. I think the server must be properly calculating a time to avoid the Thundering Herd situation.

Also, the retry-after time is returned for other services like getting thumbnail representations. If we wait for the longer time, the thumbnail representation generation will be slower.

boxsdk/session/session.py
```diff
@@ -429,11 +435,15 @@ def _get_retry_request_callable(self, network_response, attempt_number, request)
             `callable`
         """
         # pylint:disable=unused-argument
+        data = kwargs.get('data', {})
+        grant_type = None
+        if 'grant_type' in data:
```
anpr: I get the following error:
TypeError: a bytes-like object is required, not 'str'

The problem seems to be that `'grant_type'` is a `str`, but data is a bytes-object. So this can't work.
Can you confirm?

PR #481, opened 2020-01-15
Rename upload method parameters to align with API

ben-reilly:
Small iteration on #473; changing file attribute parameters to be the same as those in the API, as there's no good reason to differ from them.

### DIFF
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -200,7 +200,7 @@ class File(Item):
             preflight_expected_size=0,
             upload_using_accelerator=False,
             file_name=None,
-            file_modified_at=None,
+            content_modified_at=None,
             additional_attributes=None,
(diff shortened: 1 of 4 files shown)


boxcla:

ben-reilly:
Note that this _would_ be a breaking change, but the code I'm modifying hasn't been released yet.

PJSimon:
> Note that this _would_ be a breaking change, but the code I'm modifying hasn't been released yet.

@ben-reilly Because this PR #481 modifies code that was introduced in a previous PR, and that previous PR hasn't been released yet?  Just clarifying... Thanks!

ben-reilly:
@PJSimon - Sujay already merged. :) But yes, this is correction to #473, which was merged this week but has not been publicly released yet.

PR #484, opened 2020-01-21
Update task docs to include sample variant

PreciselyAlyss:


boxcla:
Verified that @PreciselyAlyss has signed the CLA. Thanks for the pull request!

docs/usage/task.md
```diff
@@ -177,14 +177,21 @@ To update a task assignment object, call [`assignment.update_info(data)`][update
 with a `dict` of properties to update on a task assignment. This method returns a newly update 
 [`TaskAssignment`][assignment_class] object, leaving the original object unmodified.
 
-<!-- sample put_task_assignments_id -->
+<!-- sample put_task_assignments_id resolution_state -->
```
cbetta: @PreciselyAlyss because both of these now have a variant, neither of them appear on the reference docs.

https://developer.box.com/en/reference/put-task-assignments-id/

We should probably keep one of them as the "default" variant.

Issue #483, opened 2020-01-21
TypeError during an upload_session with the chunked_uploader


anpr:
- [x] I have checked that the [SDK documentation][sdk-docs] and [API documentation][api-docs] doesn't solve my issue

### Description of the Issue
In our code, with the update of box SDK to 2.7.0 our code that uploads file with the chunked_uploader breaks, with the following exception:
`TypeError: a bytes-like object is required, not 'str'`

### Versions Used
Box Python SDK: 2.7.0
Python: 3.7.3

### Steps to Reproduce
Our code triggers the chunked upload like this:
```
chunked_uploader = BoxClient.get_chunked_uploader_for_folder(
    folder=self.folder(folder_id), file_path=file_path, file_name=file_name
)
return chunked_uploader.start()
```

### Error Message, Including Traceback
Stacktrace, up to code above (not included):
```
 /usr/local/lib/python3.7/site-packages/boxsdk/util/chunked_uploader.py(51)start()
     49         if self._is_aborted:
     50             raise BoxException('The upload has been previously aborted. Please retry upload with a new upload session.')
---> 51         self._upload()
     52         content_sha1 = self._sha1.digest()
     53         return self._upload_session.commit(content_sha1=content_sha1, parts=self._part_array)

  /usr/local/lib/python3.7/site-packages/boxsdk/util/chunked_uploader.py(107)_upload()
    105             self._sha1.update(next_part.chunk)
    106             # Retrieve the uploaded part if the part has already been uploaded. If not upload the current part.
--> 107             uploaded_part = self._part_definitions.get(next_part.offset) or next_part.upload()
    108             self._inflight_part = None
    109             # Record that the part has been uploaded.

  /usr/local/lib/python3.7/site-packages/boxsdk/util/chunked_uploader.py(193)upload()
    190         return self._upload_session.upload_part_bytes(
    191             part_bytes=self.chunk,
    192             offset=self.offset,
--> 193             total_size=self._total_size
    194         )

  /usr/local/lib/python3.7/site-packages/boxsdk/object/upload_session.py(98)upload_part_bytes()
     96             self.get_url(),
     97             headers=headers,
---> 98             data=part_bytes,
     99         )
    100         return response.json()['part']

  /usr/local/lib/python3.7/site-packages/boxsdk/session/session.py(122)put()
    120             `unicode`
    121         """
--> 122         return self.request('PUT', url, **kwargs)
    123 
    124     def delete(self, url, **kwargs):

[...]

PR #485, opened 2020-01-21
Fix for `_get_retry_request_callable` so chunked uploads continue to work

sujaygarlanka:
## Overview
Release of the Python SDK version 2.7.0 contained new logic in `_get_retry_request_callable` that prevented chunked uploads from working. 

## Details
The  new logic in the 2.7.0 release in `_get_retry_request_callable` method was reading the `data` dictionary passed in every request. However, for chunked uploads, `data` is a byte stream and not a dictionary, so the reading of the data fails. The fix in this PR is a simple try block that reads the `data` parameter when it is an object and doesn't when the `data` is a byte stream.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,11 +4,12 @@ Release History
 ---------------
 
 Next Release
-- Fixed bug in _get_retry_request_callable() introduced in release 2.7.0 which caused chunked uploads to fail
+++++++++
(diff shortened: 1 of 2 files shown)


boxcla:

PJSimon:
Resolves https://github.com/box/box-python-sdk/issues/483

PR #493, opened 2020-03-20
Add support for token exchange using shared links

boxdave:
- Adds shared_link kwarg to Client.downscope_token. If shared link is sent as an arg in the method call, the downscoped token will be scoped to the resource the shared link references.
- Adds unit test for Client.downscope_token with share_link sent as an arg, and shared_link and item not sent as an arg.

### DIFF
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -1399,7 +1399,7 @@ class Client(Cloneable):
         )
 
     @api_call
-    def downscope_token(self, scopes, item=None, additional_data=None):
+    def downscope_token(self, scopes, item=None, shared_link=None, additional_data=None):
         """
         Generate a downscoped token for the provided file or folder with the provided scopes.
 
@@ -1412,6 +1412,11 @@ class Client(Cloneable):
             not be scoped down to just a single item.
         :type item:
             :class:`Item`
+        :param shared_link:
+            (Optional) The shared link to get a downscoped token for. If None, the resulting token will
+            not be scoped down to just a single item.
+        :type item:
+            `str`
         :param additional_data:
             (Optional) Key value pairs which can be used to add/update the default data values in the request.
         :type additional_data:
@@ -1431,6 +1436,8 @@ class Client(Cloneable):
         }
         if item:
             data['resource'] = item.get_url()
+        if shared_link:
+            data['box_shared_link'] = shared_link
         if additional_data:
             data.update(additional_data)
 
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -1415,7 +1415,7 @@ class Client(Cloneable):
         :param shared_link:
             (Optional) The shared link to get a downscoped token for. If None, the resulting token will
             not be scoped down to just a single item.
-        :type item:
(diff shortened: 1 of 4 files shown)


test/unit/client/test_client.py
```diff
@@ -1072,7 +1072,7 @@ def check_downscope_token_request(
         mock_object_id,
         make_mock_box_request,
 ):
-    def do_check(item_class, scopes, additional_data, expected_data):
```
sujaygarlanka: Why did you change the order of these parameters?

boxsdk/client/client.py
```diff
@@ -1399,7 +1398,7 @@ def get_pending_collaborations(self, limit=None, offset=None, fields=None):
         )
 
     @api_call
-    def downscope_token(self, scopes, item=None, additional_data=None):
+    def downscope_token(self, scopes, item=None, shared_link=None, additional_data=None):
```
sujaygarlanka: Also, the shared_link parameter should be after additional_data to avoid a breaking change.

sujaygarlanka:
@boxdave I updated the HISTORY.rst with this change. For future fixes in the SDKs, do you mind updating the change log. In all other SDKs other than this one, it is the CHANGELOG.md file. This is something new are trying to adhere to in order to make our releases quicker.

test/unit/client/test_client.py
boxdave: I changed the order here to match the signature of Client.downscope_token. Minor thing, but in that method, scopes comes first then the item. I thought it seems a little more consistent to match that signature in do_check

test/unit/client/test_client.py
```diff
@@ -1123,7 +1120,35 @@ def test_downscope_token_sends_downscope_request(
         'scope': expected_scopes,
         'grant_type': 'urn:ietf:params:oauth:grant-type:token-exchange',
     }
-    check_downscope_token_request(item_class, scopes, {}, expected_data)
+    check_downscope_token_request(scopes, item_class, None, None, expected_data)
+
+
+def test_downscope_token_sends_downscope_request_with_no_item_shared_link_or_additional_data(
```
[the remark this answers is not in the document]
dtrodger: @sujaygarlanka It looks like this test is actually handled by the last pytest.mark.parametrize in test_downscope_token_sends_downscope_request already. Should I just remove this redundant test?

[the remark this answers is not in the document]
sujaygarlanka: You're right. I would remove this test then.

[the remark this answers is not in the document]
dtrodger: Just pushed a commit that removes this.

boxsdk/client/client.py
```diff
@@ -15,6 +14,7 @@
 from ..pagination.limit_offset_based_object_collection import LimitOffsetBasedObjectCollection
 from ..pagination.marker_based_object_collection import MarkerBasedObjectCollection
 from ..util.shared_link import get_shared_link_header
+from ..object.item import Item
```
sujaygarlanka: Can this be removed? Linter is not passing because this is imported, but not used.

PR #497, opened 2020-04-22
Add the ability to pass in a SHA1 value for file uploads

sujaygarlanka:
Added the ability to pass in a SHA1 value for file uploads. It will sent in the headers to the API with the following key `Content-MD5`.

### DIFF
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -356,7 +356,7 @@ class File(Item):
         :type additional_attributes:
             `dict` or None
         :param sha1:
-            A sha1 checksum for a file.
+            A sha1 checksum for the new content.
         :type sha1:
             `unicode` or None
         :returns:
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -202,6 +202,7 @@ class File(Item):
             file_name=None,
             content_modified_at=None,
             additional_attributes=None,
+            sha1=None
     ):
         """
         Upload a new version of a file, taking the contents from the given file stream.
@@ -244,6 +245,10 @@ class File(Item):
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A content MD5 checksum for a file.
+        :type sha1:
+            `unicode` or None
         :returns:
             A new file object
         :rtype:
@@ -275,7 +280,13 @@ class File(Item):
 
         data = {'attributes': json.dumps(attributes)}
         files = {'file': ('unused', file_stream)}
-        headers = {'If-Match': etag} if etag is not None else None
+        headers = {}
+        if etag is not None:
+            headers['If-Match'] = etag
+        if sha1 is not None:
+            headers['Content-MD5'] = sha1
+        if not headers:
+            headers = None
         file_response = self._session.post(
             url,
             expect_json_response=False,
@@ -301,6 +312,7 @@ class File(Item):
             file_name=None,
             content_modified_at=None,
             additional_attributes=None,
+            sha1=None
     ):
         """Upload a new version of a file. The contents are taken from the given file path.
 
@@ -342,6 +354,10 @@ class File(Item):
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A content MD5 checksum for a file.
+        :type sha1:
+            `unicode` or None
         :returns:
             A new file object
         :rtype:
@@ -360,6 +376,7 @@ class File(Item):
                 file_name=file_name,
                 content_modified_at=content_modified_at,
                 additional_attributes=additional_attributes,
+                sha1=sha1
             )
 
     @api_call
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -246,7 +246,7 @@ class File(Item):
         :type additional_attributes:
             `dict` or None
         :param sha1:
-            A sha1 checksum for a file.
+            A sha1 checksum for the new content.
         :type sha1:
             `unicode` or None
         :returns:
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -202,7 +202,7 @@ class File(Item):
             file_name=None,
             content_modified_at=None,
             additional_attributes=None,
-            sha1=None
+            sha1=None,
     ):
         """
         Upload a new version of a file, taking the contents from the given file stream.
@@ -313,7 +313,7 @@ class File(Item):
             file_name=None,
             content_modified_at=None,
             additional_attributes=None,
-            sha1=None
+            sha1=None,
     ):
         """Upload a new version of a file. The contents are taken from the given file path.
 
@@ -377,7 +377,7 @@ class File(Item):
                 file_name=file_name,
                 content_modified_at=content_modified_at,
                 additional_attributes=additional_attributes,
-                sha1=sha1
+                sha1=sha1,
             )
 
     @api_call
--- a/boxsdk/object/file.py
+++ b/boxsdk/object/file.py
@@ -246,7 +246,7 @@ class File(Item):
         :type additional_attributes:
             `dict` or None
         :param sha1:
-            A content MD5 checksum for a file.
+            A sha1 checksum for a file.
         :type sha1:
             `unicode` or None
         :returns:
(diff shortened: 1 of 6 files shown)


boxsdk/object/file.py
```diff
@@ -244,6 +245,10 @@ def update_contents_with_stream(
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A content MD5 checksum for a file.
```
kingsawyer, line 249: this should be sha1

boxsdk/object/file.py
```diff
@@ -275,7 +280,13 @@ def update_contents_with_stream(
 
         data = {'attributes': json.dumps(attributes)}
         files = {'file': ('unused', file_stream)}
-        headers = {'If-Match': etag} if etag is not None else None
+        headers = {}
+        if etag is not None:
+            headers['If-Match'] = etag
+        if sha1 is not None:
+            headers['Content-MD5'] = sha1
```
kingsawyer, line 287: Add comment that this is not a typo, the content-md5 field accepts sha1 hashes

boxsdk/object/file.py
```diff
@@ -342,6 +355,10 @@ def update_contents(
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A sha1 checksum for a file.
```
kingsawyer, line 359: should this say for "the new content" (different from the sha1 in the if-match which was the old content)

boxsdk/object/folder.py
```diff
@@ -306,6 +308,14 @@ def upload_stream(
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A sha1 checksum for a file.
```
kingsawyer, line 312: "for the file"

boxsdk/object/folder.py
```diff
@@ -404,6 +424,14 @@ def upload(
...
```
kingsawyer, line 428: "for the new content"

PJSimon, line 428:
```suggestion
            A sha1 checksum for the new content.
```

boxsdk/object/folder.py
PJSimon, line 312:
```suggestion
            A sha1 checksum for the file.
```

boxsdk/object/file.py
PJSimon:
```suggestion
            A sha1 checksum for the new content.
```

test/unit/object/test_file.py
```diff
@@ -310,6 +311,7 @@ def test_update_contents(
             file_name=file_new_name,
             content_modified_at=content_modified_at,
             additional_attributes=additional_attributes,
+            sha1=sha1
```
anthonywee, line 314: nit: trailing comma

test/unit/object/test_file.py
```diff
@@ -322,6 +324,7 @@ def test_update_contents(
                 file_name=file_new_name,
                 content_modified_at=content_modified_at,
                 additional_attributes=additional_attributes,
+                sha1=sha1
```
anthonywee, line 327: nit: trailing comma

boxsdk/object/file.py
```diff
@@ -360,6 +377,7 @@ def update_contents(
                 file_name=file_name,
                 content_modified_at=content_modified_at,
                 additional_attributes=additional_attributes,
+                sha1=sha1
```
anthonywee, line 380: nit: trailing comma

boxsdk/object/file.py
```diff
@@ -301,6 +313,7 @@ def update_contents(
             file_name=None,
             content_modified_at=None,
             additional_attributes=None,
+            sha1=None
```
anthonywee, line 316: nit: trailing comma

boxsdk/object/file.py
```diff
@@ -202,6 +202,7 @@ def update_contents_with_stream(
...
```
anthonywee, line 205: nit: trailing comma

boxsdk/object/file.py
```diff
@@ -244,6 +245,10 @@ def update_contents_with_stream(
             A dictionary containing attributes to add to the file that are not covered by other parameters.
         :type additional_attributes:
             `dict` or None
+        :param sha1:
+            A sha1 checksum for a file.
```
anthonywee, line 249:
```suggestion
            A sha1 checksum for the new content.
```

boxsdk/object/file.py
```diff
@@ -275,7 +280,14 @@ def update_contents_with_stream(
 
         data = {'attributes': json.dumps(attributes)}
         files = {'file': ('unused', file_stream)}
-        headers = {'If-Match': etag} if etag is not None else None
+        headers = {}
```
anthonywee, line 283: Did we consider putting this in item.py, so it can be shared between file/folder?

boxsdk/object/file.py
```diff
@@ -275,7 +280,14 @@ def update_contents_with_stream(
...
+        if etag is not None:
+            headers['If-Match'] = etag
+        if sha1 is not None:
+            # The Content-MD5 field accepts sha1
+            headers['Content-MD5'] = sha1
+        if not headers:
+            headers = None
```
anthonywee, line 290: If we make this into a helper method, we could replace these two lines with `return headers or None`

test/unit/object/conftest.py
```diff
@@ -271,11 +271,28 @@ def if_none_match_header(etag):
...
+@pytest.fixture()
+def if_match_sha1_header(etag, sha1):
+    headers = {}
+    if etag is not None:
+        headers['If-Match'] = etag
+    if sha1 is not None:
+        headers['Content-MD5'] = sha1
+    if not headers:
+        headers = None
+    return headers
```
anthonywee, line 283: ditto from above, we could do `return headers or None` and skip line 281/282

boxsdk/object/file.py
sujaygarlanka: We feel that a helper method for just three instances of creating headers may be less readable than simply including the header logic multiple times. Also, setting the headers in the method is how it is currently handled in other methods in the SDK.

boxsdk/object/file.py
sujaygarlanka: Thanks for the suggestion. After looking into this, we feel like it would be better to not consolidate the methods in `item.py`. 

First, to avoid breaking changes, we would have to keep the four methods because they are named differently. It would be nice to consolidate the upload logic since a lot of it is the same, however the  `update_contents_with_stream` in file and `upload_with_stream` in folder do slightly different things. The file method is for uploading a new file version and the folder method is for uploading a new file. Consolidating the logic would require doing some conditional logic in a single method that is part of `item.py` to change between these two calls. This extra logic may just make the SDK a little more confusing.

anthonywee: Sorry, I was actually just referring to the new code that we've added to generate the headers, since we've duplicated that in both file/folder. Was thinking something like `_get_headers_for_upload(etag, sha1)`?

test/unit/object/conftest.py
anthonywee: I'm okay to duplicate this for the tests, was thinking it could be nice to share between file/folder

boxsdk/object/file.py
anthonywee: Ah okay looks like you've discussed this idea too in the other comment below

boxsdk/object/file.py
anthonywee: Was thinking it could be nice to share this for file/folder. If we have more headers to add in the future for upload we then have a shared place to update.

Yea it's not a big deal overall so either way is okay, thanks for considering it!

test/unit/object/conftest.py
```diff
@@ -271,11 +271,28 @@ def if_none_match_header(etag):
     return {'If-None-Match': etag} if etag is not None else None
 
 
+@pytest.fixture()
+def if_match_sha1_header(etag, sha1):
```
kingsawyer, line 275: lol, of course we factored this code out in the test code, not production code.

sujaygarlanka: Included it here because it is consistent with how the tests abstract header creation.

PR #506, opened 2020-04-30
adds item id sanitization and passing test

boxdave:


boxsdk/object/item.py
```diff
@@ -725,3 +744,20 @@ def remove_classification(self):
             `bool`
         """
         return self.metadata('enterprise', self._classification_template_key).delete()
+
+    @staticmethod
+    def validate_item_id(item_id):
+        """
+        Validates an item ID is a digit
```
kingsawyer, line 751: shouldn't this be numeric, not a single digit? Otherwise we are only able to have 10 item_ids.

boxsdk/object/item.py
```diff
@@ -725,3 +744,20 @@ def remove_classification(self):
...
+
+        :param item_id:
+        :type classification:
```
kingsawyer, line 754: should be "type item_id:"

boxsdk/object/item.py
```diff
@@ -725,3 +744,20 @@ def remove_classification(self):
...
+            `str`
+        :raises:
+            BoxException: if item_id is not a digit
```
kingsawyer, line 757: "is not a string of digits" or "is not numeric"

boxsdk/object/item.py
```diff
@@ -725,3 +744,20 @@ def remove_classification(self):
...
+        :returns:
+        :rtype:
+            None
+        """
+        if not item_id.isdigit():
+            raise BoxException("Invalid item ID")
```
kingsawyer, line 763: Anything more precise we can use like ValueError?

boxsdk/object/item.py
PJSimon:
```suggestion
        Validates an item ID is numeric
```

PJSimon: @kingsawyer Great comment, thanks!  FYI for comments like this, a "GitHub Suggestion Comment" is very helpful because it is so easy to commit!

boxsdk/object/item.py
dtrodger: Good point. What do you think about adding a BoxValueError to boxsdk.exception that subclasses ValueError and raising that?

PR #516, opened 2020-06-16
Fix typo in exception variable

danlamanna:


sujaygarlanka:
@danlamanna Thank you for the PR. We really appreciate you taking an interest in help us make our SDKs better!

PR #518, opened 2020-06-17
Prep for release of v.2.9.0

PJSimon:


HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ---------------
 
+2.9.0 (2020-06-17)
+++++++++
+- Fix exception handling for OAuth
+- Documentation updates
```
sujaygarlanka, line 9: I think we can remove this and fix exception handling for OAuth. Not a bug fix or feature add users need to be aware of.

PJSimon, line 9:
```suggestion
```

PJSimon: I removed the documentation one and left the other one.  Since the exception being thrown has changed I thought there could possibly be an impact to users.

sujaygarlanka: The exception being thrown isn’t changing. Just whatever we name it in the SDK. I don’t believe the user will see the impact

PR #525, opened 2020-07-06
Fix deprecation warnings for collections.abc imports

### LINKED ISSUE #445: Deprecation warnings for collections.abc imports
# Description of the Issue

You are probably aware, but a head's up that when I run `pytest` I am getting three similar deprecation warnings:

> DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

These imports should be updated:
https://docs.python.org/3.8/library/collections.abc.html#collections.abc.Mapping

The note in 3.7 docs:
> Changed in version 3.3: Moved Collections Abstract Base Classes to the `collections.abc` module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

`boxsdk/util/log.py:5`
    `from collections import Mapping`

`boxsdk/pagination/page.py:5`
    `from collections import Sequence`

`boxsdk/pagination/box_object_collection.py:14`
    `class BoxObjectCollection(collections.Iterator, object):`

### Versions Used
Box Python SDK: 2.6.0
Python: 3.7.2

### Steps to Reproduce

Write a test that imports boxsdk

adamchainz:
Fixes #445.

### DIFF
--- a/boxsdk/object/api_json_object.py
+++ b/boxsdk/object/api_json_object.py
@@ -2,11 +2,16 @@
 
 from __future__ import unicode_literals, absolute_import
 
-from collections import Mapping
+import sys
 from abc import ABCMeta
 
 from .base_api_json_object import BaseAPIJSONObject, BaseAPIJSONObjectMeta
 from ..util.compat import with_metaclass
+
+if sys.version_info >= (3, 3):
+    from collections.abc import Mapping
+else:
+    from collections import Mapping
 
 
 class APIJSONObjectMeta(BaseAPIJSONObjectMeta, ABCMeta):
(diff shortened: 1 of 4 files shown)


boxsdk/object/api_json_object.py
```diff
@@ -2,12 +2,17 @@
...
 from __future__ import unicode_literals, absolute_import
 
-from collections import Mapping
+import sys
 from abc import ABCMeta
 
 from .base_api_json_object import BaseAPIJSONObject, BaseAPIJSONObjectMeta
 from ..util.compat import with_metaclass
 
+if sys.version_info >= (3, 3):
```
adamchainz, line 11: Using this style rather than `try/except` due to the reasoning I outlined in this post: https://adamj.eu/tech/2020/01/07/feature-checking-versus-version-checking/

PJSimon:
@adamchainz Thanks for the contribution!  We'll review this ASAP!

sujaygarlanka:
@adamchainz Why did you remove object from the Page and BoxObjectCollection classes?

adamchainz:
It's already in the MRO

```
In [3]: Iterator.__mro__
Out[3]: (collections.abc.Iterator, collections.abc.Iterable, object)
```

PR #539, opened 2020-09-15
Add zip functionality

sujaygarlanka:
Added the ability to create a zip file in Box and download it. Wrote unit tests and manually tested it.


boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        :type items:
+            `Iterable`
+        :returns:
+            A dictionary representing a created zip
+        :rtype:
+            :class:`dict`
+        """
+        # pylint: disable=protected-access
+        url = self._session.get_url('zip_downloads')
+        data_items = []
```
PJSimon, line 1690:
```suggestion
        zip_file_items = []
```

boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        for item in items:
+            data_items.append({'type': item._item_type, 'id': item.object_id})
```
PJSimon, line 1692:
```suggestion
            zip_file_items.append({'type': item._item_type, 'id': item.object_id})
```

boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        data = {
+            'download_file_name': name,
+            'items': data_items
```
PJSimon, line 1695:
```suggestion
            'items': zip_file_items
```

boxsdk/client/client.py
PJSimon: Between `data_items`, `data = {`, and `data=joson.dumps(data))` I thought renaming this variable would cut down on the usage of the word `data` here :^)

boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        :param writeable_stream:
+            Stream to pipe the readable stream of the zip file.
+        :type writeable_stream:
+            `zip`
+        :returns:
+            A status response object
+        :rtype:
+            :class:`dict`
+        """
+        created_zip = self.create_zip(name, items)
```
PJSimon, line 1725: This method always calls create_zip.  So, since there's no way for the client to download an already-created zip, then maybe create_zip should be made private?  In that case, maybe rename this method to `download_items_as_zip_file` to be more explicit?

boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        response = self._session.post(url, data=json.dumps(data)).json()
+        return self.translator.translate(
+            session=self._session,
+            response_object=response,
+        )
+
+    @api_call
+    def download_zip(self, name, items, writeable_stream):
+        """
+        Creates a zip of multiple files and folders and downloads it.
```
PJSimon, line 1706:
```suggestion
        Downloads a zip file containing multiple files and/or folders.
```

PJSimon: I think "create a zip" could be misleading because it might not be clear that the zip file is created "on Box."  Also, does the client really care that there is a creation step happening in the method?  I think they just care about the downloading of the zip file, so I removed the "creation part."

docs/usage/zip.md
```diff
@@ -0,0 +1,47 @@
+Zip
+========
+
+Allows you to create a temporary zip file of Box files and folders and download them.
```
PJSimon, line 4:
```suggestion
Allows you to create a temporary zip file on Box, containing Box files and folders, and then download them.
```

docs/usage/zip.md
```diff
@@ -0,0 +1,47 @@
...
+
+- [Create a Zip File](#create-a-zip-file)
+- [Download a Zip File](#download-a-zip-file)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+Create a Zip File
+-----------------------------
+
+Calling [`client.create_zip(name, items)`][create_zip] will let you create a new zip file with the specified name and with the specified items and will return a `dict` with the download and status link. This file does not show up in your Box account, but will be temporarily available for download.
```
PJSimon, line 18:
```suggestion
Calling [`client.create_zip(name, items)`][create_zip] will let you create a new zip file with the specified file name and containing the specified files and/or folders.  It will return a `dict` containing the download link and status link. This zip file will not show up in your Box account, but will be temporarily available for download.
```

test/unit/client/test_client.py
```diff
@@ -1464,3 +1472,111 @@ def test_device_pinner(mock_client):
 
     assert isinstance(pin, DevicePinner)
     assert pin.object_id == pin_id
+
+
+def test_create_zip(mock_client, mock_box_session):
```
PJSimon, line 1477: Do you need this test if `test_download_zip` calls `create_zip`?

boxsdk/client/client.py
sujaygarlanka: Create a zip is a public endpoint that returns information like name conflicts when a zip is created. Some users may want that information, so I think it makes sense to still keep that public.

PJSimon: But how would they download the file that was created?

test/unit/client/test_client.py
sujaygarlanka: Don't need it, but just another test because it is a separate method.

boxsdk/client/client.py
sujaygarlanka: They would get the download url from the response of create zip and make a manual api call

boxsdk/client/client.py
```diff
@@ -1666,3 +1666,68 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
...
+        :param name:
+            The name of the zip file to be created.
+        :type name:
+            `unicode`
+        :param items:
+            List of files or folders to be part of the created zip.
+        :type items:
+            `Iterable`
+        :returns:
+            A dictionary representing a created zip
```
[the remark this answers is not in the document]
PJSimon: @sujaygarlanka Are you OK with committing this suggestion?

PJSimon (approved):

PR #546, opened 2020-09-22
Add support for `copyInstanceOnItemCopy` field for metadata templates

sujaygarlanka:


boxsdk/client/client.py
```diff
@@ -1649,13 +1649,18 @@ def create_metadata_template(self, display_name, fields, template_key=None, hidd
             The scope the template should be created in
         :type scope:
             `unicode`
+        :param copy_instance_on_item_copy:
+            Tells the copy operation whether or not to copy the metadata along with the item.
```
PJSimon, line 1653:
```suggestion
            Whether or not to include the metadata when a file or folder is copied.
```

PJSimon: I created a new [issue](https://github.com/box/box-openapi/issues/118) to update developer docs to this wording, as well.

PJSimon:
@sujaygarlanka It looks like test_create_metadata_template needs to be updated.

PJSimon (approved):

PR #565, opened 2020-10-13
Fix HISTORY links to use correct repo

adamchainz:
Some copypasta had them using the ios repo, rather than Python. I was confused for a little bit trying to understand why there was so much C in the python sdk.


sujaygarlanka (approved):

sujaygarlanka:
@adamchainz Thank you for the PR! We really appreciate it.

PR #582, opened 2021-03-16
Add 'include recent shared links' support

PJSimon:


HISTORY.rst
```diff
@@ -9,6 +9,7 @@ Next Release
 **New Features and Enhancements:**
 
 - Add metadata query functionality (`#574 <https://github.com/box/box-python-sdk/pull/574>`_)
+- Add search query support for the `include_recent_shared_links` field  (`#582 <https://github.com/box/box-python-sdk/pull/574>`_)
```
swfree, line 12: url should end with`/pull/582` instead of `/pull/574`

boxsdk/object/search.py
```diff
@@ -266,6 +267,12 @@ def query(
             The direction to display the sorted search results. Can be set to `DESC` for descending or `ASC` for ascending.
         :type direction:
             `unicode` or None
+        :param include_recent_shared_links:
+            If False, the search results will only include items that are either owned by the user or items that the
+            user has been explicitly collaborated on.
+            If True, the search results may also include items that are only accessible via recently used shared links.
```
swfree, line 273: Should we also mention that the results will be returned with a slightly different format when this param is True?

sujaygarlanka (approved):

PR #585, opened 2021-03-18
Fix Travis CI

sujaygarlanka:


.travis.yml
```diff
@@ -31,7 +31,7 @@ matrix:
       dist: xenial
       sudo: true
     - python: pypy
-      env: TOX_ENV=pypy PYPY_VERSION='2.7-5.10.0'
+      env: TOX_ENV=pypy PYPY_VERSION='3.6-7.3.0'
```
swfree, line 34: do you foresee any issues with only supporting 3.6+ here? do we not need to be testing v2.7+ here?

sujaygarlanka: We are planning on deprecating 2.7 soon for the SDK since python 2.7 has been deprecated worldwide (https://www.python.org/doc/sunset-python-2/).

swfree (approved):

PR #586, opened 2021-03-22
Update `get_groups()` to use documented parameter to filter by name

sujaygarlanka:


boxsdk/client/client.py
```diff
@@ -622,7 +622,7 @@ def get_groups(self, name=None, limit=None, offset=None, fields=None):
         url = self.get_url('groups')
         additional_params = {}
         if name:
-            additional_params['name'] = name
+            additional_params['filter_term'] = name
```
swfree, line 625: any tests to add or update for this change?

swfree (approved):

PR #605, opened 2021-06-17
Fix `shared-link` docs

mwwoda:


sujaygarlanka (approved):

mwwoda:
Hi @bipsterbot 
Thanks for reaching out!
I checked it and it looks like getting the shared link require write access too, regardless if you call the API through this SDK or directly.
Thanks,
@mwwoda

PR #609, opened 2021-07-12
Remove or deprecate insensitive language

mwwoda:


boxsdk/util/translator.py
```diff
@@ -161,10 +161,10 @@ def translate(self, session, response_object):
         translated_obj = {}
         object_type = response_object.get('type', None)
         object_class = self.get(object_type) if object_type is not None else None
-        # Parent classes have the ability to "blacklist" fields that they do not want translated
-        blacklisted_fields = object_class.untranslated_fields() if object_class is not None else ()
+        # Parent classes have the ability to "denylist" fields that they do not want translated
```
PJSimon, line 164:
```suggestion
        # Parent classes have the ability to skip fields that they do not want translated
```

boxsdk/util/translator.py
```diff
@@ -161,10 +161,10 @@ def translate(self, session, response_object):
...
+        denylisted_fields = object_class.untranslated_fields() if object_class is not None else ()
```
PJSimon, line 165:
```suggestion
        fields_to_skip = object_class.untranslated_fields() if object_class is not None else ()
```

boxsdk/util/translator.py
```diff
@@ -161,10 +161,10 @@ def translate(self, session, response_object):
...
         for key in response_object:
-            if key in blacklisted_fields:
+            if key in denylisted_fields:
```
PJSimon, line 167:
```suggestion
            if key in fields_to_skip:
```

PJSimon:
Can you please rename docs/usage/(deprecated) collaboration_whitelist.md to docs/usage/collaboration_whitelist (deprecated).md so that it appears in the folder in the same alphabetical order?

test/unit/object/conftest.py
```diff
@@ -116,6 +119,21 @@ def test_collaboration_whitelist_exemption(mock_box_session, mock_object_id):
     return CollaborationWhitelistExemptTarget(mock_box_session, mock_object_id)
```
PJSimon, line 119: Should we remove: 
- test_collaboration_whitelist 
- test_collaboration_whitelist_entry
- test_collaboration_whitelist_exemption

since the related classes are now deprecated and no longer going to be updated? In other words, these tests could fail at some point in the future, but we wouldn't want to update the to fix it, right?

After you remove these three, just do a quick check to make sure they weren't being called from elsewhere in the  code.

test/unit/client/test_client.py
```diff
@@ -567,6 +568,11 @@ def test_collaboration_whitelist_initializer(mock_client):
     assert isinstance(collaboration_whitelist, CollaborationWhitelist)
```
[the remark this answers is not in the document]
mwwoda: The code coverage will drop a little, but if we don't want to support whitelist anymore I guess that's fine.

test/unit/object/conftest.py
mwwoda: Sure, same as `test_collaboration_whitelist_initializer `

test/unit/client/test_client.py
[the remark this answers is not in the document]
mwwoda: We probably should define what we mean by deprecated. Do we still support deprecated methods in some way (e.g. tests), making an effort to ensure that they are still working or do we drop the support for them altogether, leaving them just for backward compatibility - thus suffering a bit of code coverage loss. @mhagmajer what are you thoughts on this topic?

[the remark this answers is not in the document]
mhagmajer: I agree with @mwwoda. It also makes more sense to me to not make effort to cover the deprecated methods in separate tests but focus on the new ones which apps should use now.

PJSimon (approved):

test/unit/object/conftest.py
PJSimon: Decision: keep tests for deprecated methods until they break, then investigate removing both the tests and potentially EOL'ing the deprecated methods, too.

test/unit/client/test_client.py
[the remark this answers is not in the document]
PJSimon: Decision: keep tests for deprecated methods until they break, then investigate removing both the tests and potentially EOL'ing the deprecated methods, too.  Thanks!

PR #616, opened 2021-08-20
Fix PyPy build

mwwoda:
Fix related to the pyenv change. Described [here](pyenv/pyenv#1906).

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -74,4 +74,3 @@ fi
 
 source $PWD/.venv/bin/activate
 pip install -U tox
-python --version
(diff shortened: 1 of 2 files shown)


PJSimon (approved):
Shall we add something like a link to https://github.com/pyenv/pyenv/issues/1906 as a reference in the Description of this PR?  (Is that the best link?)

mwwoda:
@PJSimon 
> Shall we add something like a link to [pyenv/pyenv#1906](https://github.com/pyenv/pyenv/issues/1906) as a reference in the Description of this PR? (Is that the best link?)

Good idea!

PR #617, opened 2021-08-27
Add BoxSign support

mwwoda:
Included 2 changes to the infra. 

First one for handling '-' sign in type name. For some reason SignRequest objects use dashes instead of underscores in type field.

Second one for not retrying request on 202 code in `resend()` call.


mhagmajer (approved):
lgtm 👍

PR #619, opened 2021-09-03
Fix flaky test

mwwoda:
Dictionaries are unordered on Python < 3.6 that could sometimes cause this test to fail. Comparing the contents of the dictionaries without comparing the order of pairs solves this issue.


test/unit/client/test_client.py
```diff
@@ -646,7 +646,8 @@ def test_create_legal_hold_policy_returns_the_correct_policy_object(mock_client,
     create_policy_response.json.return_value.update(params)
     mock_box_session.post.return_value = create_policy_response
     new_policy = mock_client.create_legal_hold_policy(test_policy_name, **params)
-    mock_box_session.post.assert_called_once_with(expected_url, data=value)
+    mock_box_session.post.assert_called_once_with(expected_url, data=ANY)
+    assert dict(json.loads(mock_box_session.post.call_args[1]['data'])) == dict(json.loads(value))
```
mhagmajer, line 650: why do we need to call `dict()` on the result?

mhagmajer (approved):
Good catch!

test/unit/client/test_client.py
mwwoda: I'm packing the json to the dictionary for comparison as the result and expected values could have different order of keys/values.

PR #638, opened 2021-12-02
chore: enforce conventional commits

mwwoda:


.github/semantic.yml
```diff
@@ -0,0 +1 @@
+titleOnly: true
```
lukaszsocha2, line 1: no new line

lukaszsocha2: So if there will be anything in commit message body it will fail ?

mwwoda: Added

lukaszsocha2 (approved):
Looks good

PR #639, opened 2021-12-02
feat: Support base item operations for WebLink

lukaszsocha2:
Closes: SDK-1862


lukaszsocha2:
Items new class structure is presented on this diagram:
image

antusus (commented):
I like all the changes, but after merging this we will have new documentation for things that were not released yet. Are we planning to do a release soon? Or maybe we should move the documentation changes to separate branch and merge it before release?

lukaszsocha2:
The plan was to make a release right after merging this, but now with failing python 2.7 build I think we have to fix it or drop python 2.7 first.

antusus (approved):

Issue #641, opened 2021-12-04
Do not attempt to download a thumbnail if the representation is an error response


DaveSawyer:
### Is your feature request related to a problem? Please describe.
When a file such as a password protected PDF is the target of a thumbnail request, the result is full representations with URL, etc. Attempts to download from these URL will fail with 404. But we already know the URL is not valid due to the status result code.

### Describe the solution you'd like
Return b'' (same as missing thumbnails) for when one cannot be generated.

PR #642, opened 2021-12-04
fix: Avoid raising 404 when file cannot generate a thumbnail

DaveSawyer:
for issue #641. For some files that cannot generate thumbnails we should not attempt to download from the url field. If we do, an application may interpret the 404 not found as similar to "try again in case the thumbnail was in the process of being generated." We can also avoid an extra network call that we know will fail.

I made this intentionally narrow (as opposed to status != 'success') because some failures might be solved by waiting and retrying or actually have a valid URL.


antusus:
Hi @DaveSawyer ,

Thanks for this contribution! We will review it ASAP!

@antusus

boxsdk/object/file.py
```diff
@@ -791,9 +791,12 @@ def get_thumbnail_representation(self, dimensions, extension='png'):
             `bytes`
         """
         rep_hints = '[{0}?dimensions={1}]'.format(extension, dimensions)
-        representation = self.get_representation_info(rep_hints)
-        if representation:
-            url = representation[0]['content']['url_template']
+        representations = self.get_representation_info(rep_hints)
+        if len(representations):
+            representation = representations[0]
+            if representation['status'].get('code') in ('error_conversion_failed', 'error_password_protected'):
```
antusus, line 797: Can you tell me where did you find information that we can return this error code? I was playing with the API and with encrypted PDF and I always got error_conversion_failed.

antusus (commented):
Looks OK. I have some questions.

boxsdk/object/file.py
DaveSawyer: ah, it was reported against a file named "Password protected.docx"

antusus (approved):

antusus:
Please wait with merging until we solve the travis build issue.

antusus (changes requested):
One thing that I overlooked - could you ad an entry to HISTORY.srt

lukaszsocha2:
Also pylint is failing with messages:
`test.unit.object.test_file:904:11: [C0122(misplaced-comparison-constant), test_get_thumbnail_representation_not_found] Comparison should be thumb == ''

test.unit.object.test_file:950:11: [C0122(misplaced-comparison-constant), test_get_thumbnail_representation_not_available] Comparison should be thumb == ''`

test/unit/object/test_file.py
```diff
@@ -901,4 +901,50 @@ def test_get_thumbnail_representation_not_found(
         headers={'X-Rep-Hints': '[{}?dimensions={}]'.format(extension, dimensions)},
         params={'fields': 'representations'},
     )
-    assert thumb == b''
+    assert b'' == thumb
```
lukaszsocha2, line 904: assert thumb == b''

test/unit/object/test_file.py
```diff
@@ -901,4 +901,50 @@ def test_get_thumbnail_representation_not_found(
...
+        dimensions=dimensions,
+        extension=extension,
+    )
+
+    mock_box_session.get.assert_any_call(
+        representation_url,
+        headers={'X-Rep-Hints': '[{}?dimensions={}]'.format(extension, dimensions)},
+        params={'fields': 'representations'},
+    )
+    assert b'' == thumb
```
lukaszsocha2, line 950: assert thumb == b''

lukaszsocha2 (changes requested):

test/unit/object/test_file.py
DaveSawyer: They are equivalent and in general it's better to put a constant on the left so you don't accidentally have an assignment from a typo.

lukaszsocha2: I see your point, but according to pylint the other way around it preferred, see: https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/C0122 . Without fixing it the build will be still failing.

lukaszsocha2 (approved):

antusus (approved):

lukaszsocha2:
Hi, @DaveSawyer thanks for contributing! I took over your pr because for today a new release is planned and we wanted to include your change in it.

PR #653, opened 2021-12-31
fix: missing api_call decorators on multiput calls

DaveSawyer:
This allows callers to pass extra_network_parameters. Issue #655


boxsdk/object/upload_session.py
```diff
@@ -18,6 +19,7 @@ class UploadSession(BaseObject):
     _item_type = 'upload_session'
     _parent_item_type = 'file'
 
+    @api_call
```
lukaszsocha2, line 22: Is there any reason for adding api_call decorator to get_url() method? In other places we didn't put this decorator on this method

lukaszsocha2:
Generally looks ok, but left one question

boxsdk/object/upload_session.py
DaveSawyer: looks like this doesn't result in a network call so I removed it.

lukaszsocha2 (approved):
Looks good

PR #662, opened 2022-01-19
fix: move sphinx to test requirements

### LINKED ISSUE #661: Sphinx dependency for production code
### Description of the Issue
The 12-14 change to add Sphinx to the required packages seems like it should be a dev/test requirement only, not a production requirement. Sphinx is used to generate documentation and brings in many more requirements like babel, pytz, etc.

### Steps to Reproduce
install box sdk on a clean virtualenv. sphinx is required

### Versions Used
3.0.0

DaveSawyer:
Fixes #661. Sphinx is used to generate documentation. As such, it should be part of the dev/test requirements only. It shouldn't be required for applications that build with the SDK.


lukaszsocha2 (approved):

mwwoda:
Thanks for the contribution!

PR #689, opened 2022-02-08
feat: Add support to new python versions

lukaszsocha2:
Added support for Python 3.8, 3.9, pypy-3.7, pypy-3.8

Closes: SDK-1988

### DIFF
--- a/boxsdk/util/enum.py
+++ b/boxsdk/util/enum.py
@@ -82,8 +82,11 @@ class ExtendableEnumMeta(EnumMeta):
         return members
 
     def __contains__(cls, member):
-        if super(ExtendableEnumMeta, cls).__contains__(member):
-            return True
+        try:
+            if super(ExtendableEnumMeta, cls).__contains__(member):
+                return True
+        except TypeError:
+            return False
 
         def in_(subclass):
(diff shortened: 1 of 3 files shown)


boxsdk/util/enum.py
```diff
@@ -82,8 +82,11 @@ def __members__(cls):
         return members
 
     def __contains__(cls, member):
-        if super(ExtendableEnumMeta, cls).__contains__(member):
-            return True
+        try:
+            if super(ExtendableEnumMeta, cls).__contains__(member):
```
lukaszsocha2, line 86: This is added due to following changes in __constains__ method of Enum in Python
in Python 3.9:
`def __contains__(cls, member):
        if not isinstance(member, Enum):
            raise TypeError(
                "unsupported operand type(s) for 'in': '%s' and '%s'" % (
                    type(member).__qualname__, cls.__class__.__qualname__))
        return isinstance(member, cls) and member._name_ in cls._member_map_`

in Python 3.6:

`def __contains__(cls, member):
        return isinstance(member, cls) and member._name_ in cls._member_map_`

antusus (approved):
👍

PR #691, opened 2022-02-10
fix: Remove release-v0.2.0

adamchainz:
This looks like an accidentally committed binary?


adamchainz:
Okay, you do you :)

Use [build](https://pypi.org/project/build/) for your new release process. See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html and PEP 517.

PR #693, opened 2022-02-10
test: Remove mock dependency

adamchainz:
unittest.mock was merged into Python 3, so use that version rather than the backport package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,6 @@ def main():
     test_requires = [
         'bottle',
         'jsonpatch>1.14',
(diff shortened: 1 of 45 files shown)


adamchainz:
This has revealed some bugs in the mocks - I'll leave it for you to fix.

arjankowski:
internal ticket: SDK-2004

lukaszsocha2:
Replacing mock dependency is continued here: https://github.com/box/box-python-sdk/pull/697

PR #716, opened 2022-03-21
test: Create integration tests framework

lukaszsocha2:


antusus (commented):
I'm not sure about parametrising integration tests. This can be tested with unit test and inspecting request. If you can pass one parameter value, you should be able to pass all of them. If there is some different logic than maybe it requires separate tests. Integration tests are slow and you want to limit running time as much as possible.

test/integration_new/object/folder_itest.py
```diff
@@ -0,0 +1,215 @@
...
+
+def test_preflight_check(parent_folder):
+    file_size = 1213
+    accelerator_url = parent_folder.preflight_check(file_size, util.random_name())
+
+    assert accelerator_url
+
+
+def test_manual_chunked_upload(parent_folder, large_file_name, large_file_path):
+    total_size = os.stat(large_file_path).st_size
```
antusus, line 49: In .net and Java we are generating large file on the spot not to keep it in the source code. Maybe you can do the same. Why someone need to clone >20MB file just to run integration test?

test/integration_new/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+def small_file_v2_name():
+    return 'small_v2.pdf'
+
+
+@pytest.fixture(scope='package')
+def small_file_v2_path(small_file_v2_name):
+    return util.get_file_path(small_file_v2_name)
+
+
+@pytest.fixture(scope='package')
```
antusus, line 25: In .net and Java we are generating large file on the spot not to keep it in the source code. Maybe you can do the same. Why someone need to clone >20MB file just to run integration test?

test/integration_new/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+def large_file_name():
```
antusus, line 26: In .net and Java we are generating large file on the spot not to keep it in the source code. Maybe you can do the same. Why someone need to clone >20MB file just to run integration test?

test/integration_new/object/folder_itest.py
```diff
@@ -0,0 +1,215 @@
...
+
+    assert created_subfolder.id
+    assert created_subfolder.parent == parent_folder
+
+    util.permanently_delete(created_subfolder)
+
+
+@pytest.mark.parametrize(
+    'role', [
+        CollaborationRole.EDITOR,
```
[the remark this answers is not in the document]
lukaszsocha2: ok

test/integration_new/object/file_itest.py
```diff
@@ -0,0 +1,278 @@
...
+    test_comment = 'this is a test comment'
+    with BoxTestFile(parent_folder=parent_folder, file_path=small_file_path) as file:
+        file.add_comment(message=test_comment)
+        comment = list(file.get_comments())[0]
+
+        assert comment.id
+        assert comment.message == test_comment
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
lukaszsocha2: ok

test/integration_new/conftest.py
lukaszsocha2: done

test/integration_new/context_managers/box_retention_policy.py
```diff
@@ -0,0 +1,35 @@
...
+            name: str = DEFAULT_RETENTION_POLICY_NAME,
+            disposition_action: str = 'permanently_delete',
+            retention_length: Union[float, int] = 1
+    ):
+
+        policy_type = 'finite' if retention_length < float('inf') else 'infinite'
+        policies_found = list(CLIENT.get_retention_policies(policy_name=name, policy_type=policy_type))
+
+        if policies_found:
+            self._retention_policy = policies_found[0]
```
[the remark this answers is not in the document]
lukaszsocha2: you're right

antusus (approved):
Thanks for all the improvements

Issue #715, opened 2022-03-19
upload_session commit not returning None when there is a 202


isaacHuh:
Upload Session commit is not returning None when there is a 202
https://github.com/box/box-python-sdk/blob/44ca6db747ee246b811ed6a329f2ed63f543a32d/boxsdk/object/upload_session.py#L134
post is prob raising the exception with a 202 so this line doesn't get hit

Versions Used
Python SDK: 3.1.0
Python: 3.8.8

lukaszsocha2:
Hi @isaacHuh,
Thanks for submitting this Issue! We will take a look and get back to you ASAP!
@lukaszsocha2

lukaszsocha2:
Hi @isaacHuh,
let me know if this pr solves your issue ?
@lukaszsocha2

PR #718, opened 2022-03-23
fix: Upload session commit return None on 202

lukaszsocha2:
Closes: SDK-2089
Fixes: #715

### DIFF
--- a/boxsdk/object/upload_session.py
+++ b/boxsdk/object/upload_session.py
@@ -2,6 +2,7 @@ import base64
 import hashlib
 import json
 import os
+from http import HTTPStatus
 from typing import Any, Optional, TYPE_CHECKING, Iterable, IO
 
 from boxsdk import BoxAPIException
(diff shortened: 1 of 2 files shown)


antusus (dismissed):
👍

boxsdk/object/upload_session.py
```diff
@@ -126,13 +127,18 @@ def commit(
...
-            return None
+
+        try:
+            response = self._session.post(
+                self.get_url('commit'),
+                headers=headers,
+                data=json.dumps(body),
+            )
+        except BoxAPIException as box_api_exc:
+            if box_api_exc.status == 202:
```
kainjow, line 138: could use this from `http` module:
```suggestion
            if box_api_exc.status == HTTPStatus.ACCEPTED:
```

lukaszsocha2: good point

antusus (approved):
👍

PR #722, opened 2022-03-31
feat: Add support for multiple date time formats

lukaszsocha2:
Closes: SDK-1280


antusus (dismissed):
Just check if we should specify what time string are supported

antusus (approved):
👍

PR #726, opened 2022-04-25
test: Run integration tests on CI

lukaszsocha2:
Closes: SDK-2088


test/integration_new/__init__.py
```diff
@@ -1,18 +1,26 @@
...
+import base64
 import os
-from pathlib import Path
+import json
 
 from boxsdk.auth.jwt_auth import JWTAuth
 from boxsdk.client import Client
 
 
-def read_jwt_path_from_config(config_path: str):
```
mwwoda, line 9: Have you considered supporting .json configuration when running integration tests locally? It may be easier for some developers to download a configuration file and copy it into this particular solution solution, rather than setting an env variable encoded as base64 for the entire machine. For example, our .NET SDK looks for the env variable and, if it is not present, it takes the .json configuration found in the base dir.

mwwoda (approved):
Approved with a minor comment regarding potential improvements.

PR #760, opened 2022-09-15
fix: Do not log the content of downloaded file

lukaszsocha2:
Closes: SDK-2544

### DIFF
--- a/boxsdk/network/default_network.py
+++ b/boxsdk/network/default_network.py
@@ -2,10 +2,12 @@ from logging import getLogger
 from pprint import pformat
 import sys
 import time
-from typing import Any, Callable, Type
+from typing import Any, Callable, Type, Optional
 
 import requests
 from requests import Response
+from requests.structures import CaseInsensitiveDict
+from urllib3 import HTTPResponse
 
 from .network_interface import Network, NetworkResponse
 from ..util.log import sanitize_dictionary
@@ -33,12 +35,14 @@ class DefaultNetwork(Network):
         The logging of the response is deferred to :class:`DefaultNetworkResponse`.
         See that class's docstring for more info.
         """
+        log_response_content = kwargs.pop('log_response_content', True)
         self._log_request(method, url, **kwargs)
         # pylint:disable=abstract-class-instantiated
         try:
             return self.network_response_constructor(
                 request_response=self._session.request(method, url, **kwargs),
                 access_token_used=access_token,
+                log_response_content=log_response_content
             )
         except Exception:
             self._log_exception(method, url, sys.exc_info())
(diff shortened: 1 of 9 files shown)


boxsdk/network/default_network.py
```diff
@@ -140,30 +144,23 @@ class DefaultNetworkResponse(NetworkResponse):
...
 
-    def __init__(self, request_response: 'Response', access_token_used: str):
+    def __init__(self, request_response: 'Response', access_token_used: str, log_response_content: bool = True):
         self._logger = getLogger(__name__)
         self._request_response = request_response
         self._access_token_used = access_token_used
         self._did_log = False
-        if not self.ok:
-            self.log(can_safely_log_content=True)
+        self.log(can_safely_log_content=log_response_content)
```
Jeff-Meadows, line 154: I would strongly discourage making this change (moving the logging to the response constructor). It requires the caller of every network call to decide whether or not to log, rather than getting it right 99% or more of the time (with the current implementation).

It's probably OK to still look for the `log_response_content` in the constructor like we're doing, but save it as an instance variable, and combine it with `can_safely_log` in the `log` method. This would allow callers to opt out of logging, but would still get it right automatically most of the time.

Jeff-Meadows (commented):
I think it's fine to allow clients of the network layer to opt out of logging, but I think this change as-is has some issues.

The SDK already doesn't log the content of downloaded files (as long as the developer uses the methods intended for downloading files). Yes, it's possible to use the wrong SDK methods to download a file, and then end up logging file contents. It seems to me that we could improve the documentation to further discourage developers from doing that.

antusus (approved):
👍

boxsdk/network/default_network.py
lukaszsocha2: `It requires the caller of every network call to decide whether or not to log` - this is not true.
I used already existing parameter `expect_json_response`, which is being used each time when we aren't expecting json response (No Content response or bytes, e.g. for downloading files) and assign it to `log_response_content`. (log_response_content=request.expect_json_response). In request object the default value of `expect_json_response` is set True. Therefore if no `expect_json_response` specified in request() call content will be logged - don't need to specify anything everything is done automatically.

`It's probably OK to still look for the log_response_content in the constructor like we're doing, but save it as an instance variable, and combine it with can_safely_log in the log method. This would allow callers to opt out of logging, but would still get it right automatically most of the time.` - I don't see why this is wrong. When logging enabled I think we want log every network response so placing it in constructor makes sure that every one is logged (in opposition to previous solution when we logged only when accessing content or json, etc.). At this point we already know if we want to log response or not (We want to log messages only with json response) so why not put it there.

lukaszsocha2: `The SDK already doesn't log the content of downloaded files (as long as the developer uses the methods intended for downloading files)` but if they want to get the content of the file `file_content = client.file(file_id).content()` to do some operation on bytes first and then save it for file, then it is not possible to disable logging in current implementation.

lukaszsocha2: @Jeff-Meadows have above comments answered all your concerns? If no please let me know, otherwise I'll merge it tomorrow.

Jeff-Meadows:
👍

PR #761, opened 2022-09-21
feat: Fix closing file after chunked upload

lukaszsocha2:
Closes: SDK-2611


antusus (commented):
I'm not sure if the closing of the stream only when exception occurs will still not block the file on windows? Maybe we should close it in finally block?

antusus (approved):
👍

PR #780, opened 2022-12-21
fix: Fix index error when getting an empty list of user term of service statuses

lukaszsocha2:
Closes: SDK-2819


arjankowski (approved):
👍

Issue #776, opened 2022-12-07
trash().get_items does not support user_marker=True


emboehm:
I noticed that when I tried to use

```
trash_items = client.trash().get_items(use_marker=True)
```

I would get the following error

```
Traceback (most recent call last):
  File "./myapp_marker.py", line 21, in <module>
    trash_items = client.trash().get_items(use_marker=True)
  File "/work/box-python-sdk/boxsdk/util/api_call_decorator.py", line 63, in call
    return method(*args, **kwargs)
TypeError: get_items() got an unexpected keyword argument 'use_marker'
```

I verified that the API supports use_marker=True for trash
[https://developer.box.com/reference/get-folders-trash-items/](url)

### Steps to Reproduce
1. Create a box client object but authorizing to Box. I used OAuth2 for my example.
2. Try to run this code 'trash_items = client.trash().get_items(use_marker=True)'

### Expected Behavior
I expected to receive a list of items from the trash.

### Error Message, Including Stack Trace
```
Traceback (most recent call last):
  File "./myapp_marker.py", line 21, in <module>
    trash_items = client.trash().get_items(use_marker=True)
  File "/work/box-python-sdk/boxsdk/util/api_call_decorator.py", line 63, in call
    return method(*args, **kwargs)
TypeError: get_items() got an unexpected keyword argument 'use_marker'
```

### Possible fix
I looked at the code in
[https://box-python-sdk.readthedocs.io/en/latest/_modules/boxsdk/object/trash.html#Trash.get_items](url)
and compared it with the code for folders, that correctly implements use_marker=True
[https://box-python-sdk.readthedocs.io/en/latest/_modules/boxsdk/object/folder.html#Folder.get_items](url)

I identified the missing code and added the changes to boxsdk/object/trash.py.

```
diff --git a/boxsdk/object/trash.py b/boxsdk/object/trash.py
index 0f421c3..fe8363a 100644
--- a/boxsdk/object/trash.py
+++ b/boxsdk/object/trash.py
@@ -4,6 +4,7 @@
 from .base_endpoint import BaseEndpoint
 
 from ..pagination.limit_offset_based_object_collection import LimitOffsetBased
+from ..pagination.marker_based_object_collection import MarkerBasedObjectColle
 from ..util.api_call_decorator import api_call
 
 if TYPE_CHECKING:
@@ -91,7 +92,12 @@ def permanently_delete_item(self, item: 'BaseItem') -> bool:
         return box_response.ok
 
     @api_call
[...]

PR #781, opened 2022-12-21
feat: Add support marker in trash get items

congminh1254:
Closes: SDK-2810, #776

### DIFF
--- a/boxsdk/object/trash.py
+++ b/boxsdk/object/trash.py
@@ -127,9 +127,9 @@ class Trash(BaseEndpoint):
         additional_params = {}
         if limit is not None:
             additional_params['limit'] = limit
-        if not direction:
+        if direction:
             additional_params['direction'] = direction
-        if not sort:
+        if sort:
             additional_params['sort'] = sort
 
         if use_marker:
--- a/boxsdk/object/trash.py
+++ b/boxsdk/object/trash.py
@@ -99,10 +99,10 @@ class Trash(BaseEndpoint):
         limit: Optional[int] = None,
         offset: Optional[int] = None,
         fields: Iterable[str] = None,
-        sort: Optional[str] = None,
-        direction: Optional[str] = None,
         marker: Optional[str] = None,
-        use_marker: bool = False
+        use_marker: bool = False,
+        sort: Optional[str] = None,
+        direction: Optional[str] = None
     ) -> 'BoxObjectCollection':
         """
         Using limit-offset paging, get the files, folders and web links that are in the user's trash.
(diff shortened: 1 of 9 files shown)


boxsdk/object/trash.py
```diff
@@ -101,14 +113,43 @@ def get_items(self, limit: Optional[int] = None, offset: Optional[int] = None, f
             The offset of the item at which to begin the response.
         :param fields:
             List of fields to request.
+        :param sort:
```
lukaszsocha2, line 116: Maybe lets keep the orgde of params the same as for folder.get_items()
` marker: Optional[str] = None,
            use_marker: bool = False,
            sort: Optional[str] = None,
            direction: Optional[str] = None,`

Except `fields` of course in order not to make breaking change

boxsdk/object/trash.py
```diff
@@ -101,14 +113,43 @@ def get_items(self, limit: Optional[int] = None, offset: Optional[int] = None, f
...
+            The field to sort by.
```
lukaszsocha2, line 117: Maybe add possible values: `Value is one of id, name, date, size`

boxsdk/object/trash.py
```diff
@@ -101,14 +113,43 @@ def get_items(self, limit: Optional[int] = None, offset: Optional[int] = None, f
...
+        :param direction:
+            The direction to sort.
```
lukaszsocha2, line 119: Add `Value is one of ASC,DESC`

boxsdk/object/trash.py
```diff
@@ -101,14 +113,43 @@ def get_items(self, limit: Optional[int] = None, offset: Optional[int] = None, f
...
+            The marker at which to begin the response.
+        :param use_marker:
+            Whether or not to use marker-based paging.
         :returns:
             An iterator of the entries in the trash
         """
+        additional_params = {}
+        if limit is not None:
+            additional_params['limit'] = limit
+        if direction is not None:
```
lukaszsocha2, line 130: maybe just `if not direction:`. Empty string is also not valid in this case. Same for limit if statement

lukaszsocha2:
Maybe add integration test for getting trash items using marker pagination

test/integration_new/object/trash_itest.py
```diff
@@ -0,0 +1,77 @@
...
+        yield folder
+
+
+def test_trash_get_items(parent_folder, small_file_path):
+    name = f'{util.random_name()}.pdf'
+    test_file = parent_folder.upload(file_path=small_file_path, file_name=name)
+    test_file.delete()
+    trash_items = CLIENT.trash().get_items()
+    for item in trash_items:
+        if item.id == test_file.id:
```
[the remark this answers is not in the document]
lukaszsocha2: `assert test_file.id in [item.id for item in trash_items]`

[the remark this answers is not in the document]
lukaszsocha2: how about only such assert ? checking name and type is not essential for this test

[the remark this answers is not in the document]
lukaszsocha2: sth like this:
```
try:
  trash_items = CLIENT.trash().get_items()
  assert test_file.id in [item.id for item in trash_items]
finally:
   CLIENT.trash().permanently_delete_item(test_file)
```

lukaszsocha2 (approved):
LGTM

PR #800, opened 2023-02-13
feat: Use multiple threading for chunked upload

congminh1254:


boxsdk/util/chunked_uploader.py
```diff
@@ -77,27 +86,42 @@ def abort(self) -> bool:
...
-            # Retrieve the uploaded part if the part has already been uploaded. If not upload the current part.
-            uploaded_part = self._part_definitions.get(next_part.offset) or next_part.upload()
-            self._inflight_part = None
-            # Record that the part has been uploaded.
-            self._part_array.append(uploaded_part)
+        try:
+            for future in as_completed(futures):
+                future.result()
+        except Exception as exc:
+            self._executor.shutdown(wait=True)
```
mwwoda, line 102: Looks like by default `wait` is set to `True` - https://docs.python.org/3/library/concurrent.futures.html
You could get rid of it if you want

boxsdk/util/chunked_uploader.py
```diff
@@ -77,27 +86,42 @@ def abort(self) -> bool:
...
+            raise exc
+
+        self._part_array = sorted(self._part_definitions.values(), key=lambda part: part['offset'])
+
+    def _upload_part(self, task=None):
+        if isinstance(task, InflightPart):
+            next_part = task
```
mwwoda, line 109: I was wondering do we need to update the hash in this case?

mwwoda (commented):
Couple of nits

test/unit/object/test_chunked_upload.py
```diff
@@ -246,11 +247,82 @@ def test_resume_in_process(test_file, mock_upload_session):
...
+
+def test_resume_with_upload_random_duration(test_file, mock_upload_session):
+    # pylint: disable-msg=unused-argument
+    def upload_mock_func(part_bytes, offset, total_size):
+        value = upload_results.pop(0)
+        if isinstance(value, Exception):
+            time.sleep(randint(0, 1))
+            raise value
+        time.sleep(randint(0, 1))
+        return value
```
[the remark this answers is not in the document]
lukaszsocha2: Sure I was testing different delays for exc and actual response, but forgot to move it as I use now same value.

boxsdk/util/chunked_uploader.py
lukaszsocha2: I know, but I want to make it clear that we wait

boxsdk/util/chunked_uploader.py
lukaszsocha2: Hash for a part is updated only just after it is read from stream. If the task is `InflightPart` it means that we have already tried to upload this part, but it failed - so we now retry this part. Hash has already been updated for this one by the first attempt.

mwwoda: makes sense!

PR #802, opened 2023-02-17
fix: chunked uploader - fix resume error with `sha1` hash

### LINKED ISSUE #495: Chunked Uploader uses incorrect whole file digest upon resume

### Description of the Issue
When using the Chunked Uploader, if the entire file completes uploading with the call to 'start' method, then it works as expected. If there is a need to use the "resume" method, then the upload almost always fails at the "commit" step due to mismatch in the whole file SHA1 digest. I could figure out the cause, and have a suggested fix mentioned below.

### Versions Used
Box Python SDK: 2.7.1 (from PyPi)
Python: 3.6.5

### Steps to Reproduce
Upload a file using Python BoxSDK chunked uploader routine. If there is an interruption to the upload (for ex. network error), and I use the "resume" method, it causes the overall upload to fail due to whole file digest mismatch at the "commit" step.

### Root Cause and Suggested solution
boxsdk/util/chunked_uploader.py
Line: 105
Class: ChunkedUploader; Method: _upload

The line "self._sha1.update(next_part.chunk)" occurs before the part is uploaded. Hence if the part upload fails for any reason, when the ChunkedUploader retries that part upload, the SHA1 is updated with the same chunk a 2nd time causing the overall file digest to be incorrect.

Hence this line needs to be moved lower down in the same method after the part upload has successfully completed. 6 lines down, right after the "# Record that the part has been uploaded." line would be a good location to keep the  "self._sha1.update(next_part.chunk)" line.

I've pasted the full code of the fixed method below ...

    def _upload(self):
        """
        Utility function for looping through all parts of of the upload session and uploading them.
        """
        while len(self._part_array) < self._upload_session.total_parts:
            # Retrieve the part inflight if it exists, if it does not exist then get the next part from the stream.
            next_part = self._inflight_part or self._get_next_part()
            # Set the retrieve part to the current part inflight.
            self._inflight_part = next_part
            # Retrieve the uploaded part if the part has already been uploaded. If not upload the current part.
            uploaded_part = self._part_definitions.get(next_part.offset) or next_part.upload()
            self._inflight_part = None
            # Record that the part has been uploaded.
            self._sha1.update(next_part.chunk)
            self._part_array.append(uploaded_part)
            self._part_definitions[next_part.offset] = uploaded_part

<class 'boxsdk.exception.BoxAPIException'>.
Message: File digest was incorrect. Actual: V1UVe3W571bzWYXxlmrfDvNmexk= Expected: iHU8e+hNlA0T/ElfbfcFtf+0VdM=
Status: 400
Code: digest_mismatch
Request ID: 9ce5dd36ea3ed10821e637bb362c40df
Headers: {'Date': 'Fri, 03 Apr 2020 09:33:40 GMT', 'Content-Type': 'application/json; charset=UTF-8', 'Content-Length': '189', 'Connection': 'keep-alive', 'Strict-Transport-Security': 'max-age=31536000'}
URL: https://upload.box.com/api/2.0/files/upload_sessions/0536358B86DBFB0A2080FF2840E0038B/commit
Method: POST
Context Info: None.

[sdk-docs]: ./docs/usage
[api-docs]: https://developer.box.com/

hoggatt:
Fixes #495
when the sha1 hash is updated before the uploaded part actually succeeds and something happens that requires resuming, a digest mismatch error is thrown (see issue #495)

### DIFF
--- a/boxsdk/util/chunked_uploader.py
+++ b/boxsdk/util/chunked_uploader.py
@@ -91,11 +91,11 @@ class ChunkedUploader:
             next_part = self._inflight_part or self._get_next_part()
             # Set the retrieve part to the current part inflight.
             self._inflight_part = next_part
-            self._sha1.update(next_part.chunk)
             # Retrieve the uploaded part if the part has already been uploaded. If not upload the current part.
             uploaded_part = self._part_definitions.get(next_part.offset) or next_part.upload()
             self._inflight_part = None
             # Record that the part has been uploaded.
+            self._sha1.update(next_part.chunk)
             self._part_array.append(uploaded_part)
             self._part_definitions[next_part.offset] = uploaded_part
 


hoggatt:
Made this change locally and it ended up fixing the problem.

lukaszsocha2:
Hi @hoggatt,
thanks for contributing. In recent pr (https://github.com/box/box-python-sdk/pull/800) we have introduced a feature to use multiple threads for chunk upload. Your change was also included in that pr, so I will close this one. New feature will be available soon in new release.
Best,
@lukaszsocha2

PR #815, opened 2023-05-08
fix: Use the older version of `urllib3`

congminh1254:


tox.ini
```diff
@@ -96,7 +96,7 @@ deps =
     twine
 
 [testenv:integration-tests]
-passenv = JWT_CONFIG_BASE_64 ADMIN_USER_ID
+passenv = JWT_CONFIG_BASE_64 ADMIN_USER_ID
```
lukaszsocha2, line 99: extra space

.github/workflows/integration-tests.yml
```diff
@@ -11,17 +11,23 @@ jobs:
...
+      - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
-          python-version: '3.6'
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install tox
+          python -m pip install tox==3.28.0
```
lukaszsocha2, line 30: why was this required?

congminh1254: Hi Lukasz :smile:
about the `tox==3.28.0`
The latest version of tox doesn’t support python3.6, so if we not provide any version, python 3.6 will get tox 3.28.x and the python 3.8 will get tox 4.x
and it have difference in term of `passenv` config, in version 3x, we can use: `passenv = JWT_CONFIG_BASE_64 ADMIN_USER_ID` but in 4.x, we need to use `passenv = JWT_CONFIG_BASE_64,ADMIN_USER_ID`, kind of breaking change.
So in here, I specified that we only use 3.28.0 as tox version.

ryandeivert:
@congminh1254 any chance you can release this any time soon?

congminh1254:
Hi @ryandeivert 
We have just released the new version of the SDK, you can update it now.
Best,
Minh

PR #824, opened 2023-07-06
fix: Stop processing data for logging when logging is disabled and cache response `json`

lukaszsocha2:
Closes: SDK-3230


mwwoda (approved):
LGTM!

PR #831, opened 2023-07-27
fix: Fix JSON response validation retry strategy

congminh1254:


test/unit/session/test_session.py
```diff
@@ -221,12 +221,23 @@ def test_box_session_seeks_file_after_retry(box_session, mock_network_layer, ser
     assert mock_file_2.seek.has_calls(call(3) * 2)
 
 
-def test_box_session_raises_for_non_json_response(box_session, mock_network_layer, non_json_response, test_url):
+def test_box_session_raises_for_non_json_response(box_session, mock_network_layer, non_json_response, generic_successful_response, test_url):
     # pylint:disable=redefined-outer-name
-    mock_network_layer.request.side_effect = [non_json_response]
+    mock_network_layer.request.side_effect = [non_json_response, generic_successful_response]
 
+    API.MAX_RETRY_ATTEMPTS = 1
```
lukaszsocha2, line 228: put `API.MAX_RETRY_ATTEMPTS = 5` to finally block

lukaszsocha2, line 228: not need to modify API.MAX_RETRY_ATTEMPTS

test/unit/session/test_session.py
```diff
@@ -221,12 +221,23 @@ def test_box_session_seeks_file_after_retry(box_session, mock_network_layer, ser
...
+    box_session.get(url=test_url)
+    API.MAX_RETRY_ATTEMPTS = 5
+
+
+def test_box_session_raises_for_non_json_response_after_retry(box_session, mock_network_layer, non_json_response, test_url):
+    # pylint:disable=redefined-outer-name
+    mock_network_layer.request.side_effect = [non_json_response, non_json_response]
```
lukaszsocha2, line 235: [non_json_response] * API.MAX_RETRY_ATTEMPTS , and you don't have to modify this variable

boxsdk/session/session.py
```diff
@@ -541,6 +556,10 @@ def _get_retry_request_callable(
             self._renew_session(request.access_token)
             request.auto_session_renewal = False
             return self._send_request
+
+        if request.method == 'GET' and network_response and network_response.ok and not self._validate_json_response(network_response, request):
```
lukaszsocha2, line 560: You put this piece into AuthorizedSession class. I guess the better place would be to add it to parent class Session to overriden method, because this thing can happen both dor authorized and unauthorized session

boxsdk/session/session.py
```diff
@@ -279,15 +279,29 @@ def _raise_on_unsuccessful_request(network_response: 'NetworkResponse', request:
                 context_info=response_json.get('context_info', None),
                 network_response=network_response
             )
+
+    @staticmethod
+    def _validate_json_response(network_response: 'NetworkResponse', request: '_BoxRequest', throw_exception=False) -> bool:
```
lukaszsocha2, line 284: it is a bad practice to design methods like that with a bool flag. `_validate_json_response` doesn't indicate that a bool will be a response and that exception can be thrown. I would create sth like:
```
def _is_response_format_valid(network_response: 'NetworkResponse', request: '_BoxRequest'):
    return request.expect_json_response and not is_json_response(network_response)
```
and keep throwing error in `_raise_on_unsuccessful_request`

lukaszsocha2 (changes requested):
Pls apply comments

boxsdk/session/session.py
```diff
@@ -289,6 +290,20 @@ def _raise_on_unsuccessful_request(network_response: 'NetworkResponse', request:
...
+    def _validate_json_response(network_response: 'NetworkResponse', request: '_BoxRequest') -> bool:
+        """
+        Validate that the response is json if the request expects json response.
+
+        :param network_response:
+            The network response which is being tested for success.
+        :param request:
+            The API request that could be unsuccessful.
+        """
+        if request.expect_json_response and not is_json_response(network_response):
```
lukaszsocha2, line 303: could be simplified to simple return statement.

lukaszsocha2: if function return bool it is a good practice to name functions strating from is_...

boxsdk/session/session.py
```diff
@@ -398,6 +413,10 @@ def _get_retry_request_callable(
                 self._send_request,
             )
         code = network_response.status_code
+
+        if request.method == 'GET' and network_response and network_response.ok and not self._validate_json_response(network_response, request):
```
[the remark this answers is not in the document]
lukaszsocha2: `and network_response` is it necessary ? `if network_response is None:` case is handled above

boxsdk/session/session.py
lukaszsocha2: maybe _is_json_response_if_expected, or sth like this

boxsdk/session/session.py
```diff
@@ -398,6 +413,10 @@ def _get_retry_request_callable(
...
+            return self._send_request
```
lukaszsocha2, line 418: Partial should be used to apply exponential backoff logic

test/unit/session/test_session.py
```diff
@@ -221,12 +221,21 @@ def test_box_session_seeks_file_after_retry(box_session, mock_network_layer, ser
...
     # pylint:disable=redefined-outer-name
-    mock_network_layer.request.side_effect = [non_json_response]
+    mock_network_layer.request.side_effect = [non_json_response, generic_successful_response]
+
+    box_session.get(url=test_url)
+
+
+def test_box_session_raises_for_non_json_response_after_retry(box_session, mock_network_layer, non_json_response, test_url):
+    # pylint:disable=redefined-outer-name
+    API.MAX_RETRY_ATTEMPTS = 1
```
lukaszsocha2, line 233: NO need to modufy this variable at all now!

boxsdk/session/session.py
```diff
@@ -391,13 +404,22 @@ def _get_retry_request_callable(
...
+        # pylint:disable=line-too-long
+        if network_response is None or (network_response.ok and request.method == 'GET' and not self._is_json_response_if_expected(network_response, request)):
             return partial(
                 self._network_layer.retry_after,
                 self.get_retry_after_time(attempt_number, None),
                 self._send_request,
             )
         code = network_response.status_code
+
+        if request.method == 'GET' and network_response.ok and not self._is_json_response_if_expected(network_response, request):
```
lukaszsocha2, line 416: duplicated

lukaszsocha2 (approved):

PR #834, opened 2023-08-01
feat: Support create sign request calls with sign template IDs

dtrodger:


congminh1254 (approved):
Thanks for your contribution, moved the param to the end to avoid conflict with some existing uses.
And also update the tests.

PR #835, opened 2023-08-01
feat: adds get sign template and get sign templates methods on Client

dtrodger:
Tell @PJSimon Dave Rodgers says hi :)


congminh1254:
Hi @dtrodger,
I have just added the tests for this sign templates, but it turned out that the response body from API is missing `type` field, it could make some issue while serialise the object.
So I'm contacting with the team who own this API to see if they can add that field to the request body, then we can merge this PR.
Thanks for your patience.
Regards,
Minh

congminh1254 (approved):

PR #867, opened 2024-02-15
fix: Fix integration test

congminh1254:


setup.py
```diff
@@ -68,10 +68,10 @@ def main():
         'bottle',
         'jsonpatch>1.14',
         'sqlalchemy<1.4.0',
-        'pytest',
-        'pytest-timeout',
-        'pytest-cov',
-        'pytest-lazy-fixture',
+        'pytest<=7.4.4',
```
lukaszsocha2, line 71: all these restrictions are required ?

lukaszsocha2: add comment why

lukaszsocha2 (approved):

PR #875, opened 2024-05-23
feat: Use upload session `urls` for chunk upload

lukaszsocha2:
Closes: SDK-3836

### DIFF
--- a/boxsdk/object/upload_session.py
+++ b/boxsdk/object/upload_session.py
@@ -9,6 +9,7 @@ from boxsdk import BoxAPIException
 from boxsdk.util.api_call_decorator import api_call
 from boxsdk.util.chunked_uploader import ChunkedUploader
 from boxsdk.session.session import Session
+from boxsdk.config import API
 from .base_object import BaseObject
 from ..pagination.limit_offset_based_dict_collection import LimitOffsetBasedDictCollection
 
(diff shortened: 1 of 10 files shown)


boxsdk/object/upload_session.py
```diff
@@ -20,10 +21,20 @@ class UploadSession(BaseObject):
...
     _parent_item_type = 'file'
 
-    def get_url(self, *args: Any) -> str:
+    def __init__(
+            self, session: Session, object_id: str, response_object: dict = None, use_upload_session_urls: bool = True
+    ):
+        super().__init__(session, object_id, response_object)
+        self._use_upload_session_urls = use_upload_session_urls
+
+    def get_url(self, *args: Any, url_key: str = None) -> str:
```
congminh1254, line 30: In case of existing customer using custom `API. UPLOAD_URL` they will be use the new URL after this change?

lukaszsocha2: Yes, unless they set use_upload_session_urls=False

congminh1254: So in case you are aware of this situation, this PR is good to go :D

congminh1254 (approved):

congminh1254 (approved):
LGTM :+1:

PR #878, opened 2024-07-10
feat: add create sign request function with different required parameters

mwwoda:
Some of the params (`parent_folder_id` and `files`)  are no longer required since https://github.com/box/box-openapi/commit/797a90d21024acb03354328cd921c989fac36f4f

### DIFF
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -1512,8 +1512,8 @@ class Client(Cloneable):
         """
         return self.translator.get('sign_request')(session=self._session, object_id=sign_request_id)
 
-    def __create_sign_request(
     # pylint: disable=too-many-branches
+    def __create_sign_request(
             self,
             signers: Iterable,
             files: Optional[Iterable] = None,
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -1513,6 +1513,7 @@ class Client(Cloneable):
         return self.translator.get('sign_request')(session=self._session, object_id=sign_request_id)
 
     def __create_sign_request(
+    # pylint: disable=too-many-branches
             self,
             signers: Iterable,
             files: Optional[Iterable] = None,
--- a/boxsdk/client/client.py
+++ b/boxsdk/client/client.py
@@ -1528,7 +1528,7 @@ class Client(Cloneable):
             is_document_preparation_needed: Optional[bool] = None,
             redirect_url: Optional[str] = None,
             declined_redirect_url: Optional[str] = None,
-            template_id: Optional[str] = None) -> dict:
+            template_id: Optional[str] = None) -> SignRequest:
         url = self._session.get_url('sign_requests')
 
         body = {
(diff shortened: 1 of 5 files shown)


boxsdk/client/client.py
```diff
@@ -1570,46 +1631,71 @@ def create_sign_request(
...
+            are_text_signatures_enabled: Optional[bool] = None,
+            days_valid: Optional[str] = None,
+            email_message: Optional[Iterable] = None,
+            email_subject: Optional[str] = None,
+            external_id: Optional[str] = None,
+            is_document_preparation_needed: Optional[bool] = None,
+            redirect_url: Optional[str] = None,
+            declined_redirect_url: Optional[str] = None,
+            template_id: Optional[str] = None,
+    ) -> dict:
```
lukaszsocha2, line 1655: -> 'SignRequest'
(I know it was already like that but we can fix it ;)

lukaszsocha2: apply to all 3 functions

docs/usage/sign_requests.md
```diff
@@ -32,12 +35,23 @@ files = [source_file]
...
 signer = {
     'name': 'John Doe',
-    'email': 'signer@mail.com' 
+    'email': 'signer@mail.com'
 }
 signers = [signer]
 
-parent_folder_id = '123456789'
-new_sign_request = client.create_sign_request(files, signers, parent_folder_id)
+new_sign_request = client.create_sign_request_v2(signers, files)
```
lukaszsocha2, line 42: Why removing parent_folder_id from docs? I think we should add kwargs to optional names 
`client.create_sign_request(files, signers=signers, parent_folder_id=parent_folder_id)`

mwwoda: sure we could keep it, but it's no longer a required arg

boxsdk/client/client.py
mwwoda: sure

docs/usage/sign_requests.md
```diff
@@ -32,12 +35,23 @@ files = [source_file]
...
+
+Example with sign template
+
+```python
+signer = {
+    'name': 'John Doe',
+    'email': 'signer@mail.com'
+}
+
+new_sign_request = client.create_sign_request_v2(signers, template_id = '12345')
```
lukaszsocha2, line 54: No spaces around `template_id = '12345'`, otherwise linter will complain :D

mwwoda: Linter didn't complain about this particular line :D. But sure, I'll fix it

lukaszsocha2: Because it is in md file. If you paste it to .py file it would complain ;)

docs/usage/sign_requests.md
lukaszsocha2: I know that is why I suggested using kwargs

lukaszsocha2 (approved):
You can merge if build is green

PR #890, opened 2025-03-28
feat: Add `stream_file_content` parameter to upload methods

lukaszsocha2:


metathinker:
Hi! For my understanding, I have a couple conceptual questions about this particular change:

1. In our application, the streams we pass to the upload() calls are actual open files, so random access is possible. Could we add an option to have requests rewind the stream and reread it from the start, so that the content can be read twice _without_ having to preload it into memory? Or would that require code changes in requests itself?
2. Is it possible to make the same change as this one to chunked upload sessions, so that chunk uploads can also follow 307 redirects successfully?

lukaszsocha2:
Hi,
1. It would be possible, but we would have to disable auto redirects in `requests` library and handle all redirects manually. I'm not sure if we want to make such a big change in our SDK
2. Chunks of a file are loaded into memory already in case retry of a chunk is necessary. Did you test it and it didn't work this way?

metathinker:
>> 1. In our application, the streams we pass to the upload() calls are actual open files, so random access is possible. Could we add an option to have requests rewind the stream and reread it from the start, so that the content can be read twice without having to preload it into memory? Or would that require code changes in requests itself?
> 1. It would be possible, but we would have to disable auto redirects in `requests` library and handle all redirects manually. I'm not sure if we want to make such a big change in our SDK

That makes sense.
>> 2. Is it possible to make the same change as this one to chunked upload sessions, so that chunk uploads can also follow 307 redirects successfully?
> 2. Chunks of a file are loaded into memory already in case retry of a chunk is necessary. Did you test it and it didn't work this way?

Sorry, my mistake - please feel free to ignore my point 2.

Our application actually bypasses the [`UploadSession.upload_part_bytes()`](https://github.com/box/box-python-sdk/blob/320ffe95ce4d9dc826c14e9f075e0aedef10f685/boxsdk/object/upload_session.py#L67-L89) method, which behaves as you describe, and instead reaches into the BoxSession to call PUT at a lower level. After a closer reading, it appears to me that our app does this precisely to avoid having to pre-load the chunk into memory. We will have to undo that behavior ourselves.

lukaszsocha2:
1. When setting `stream_file_content` to False SDK rollbackS to the default `requests` library behaviour. I read in docs that it keeps body in the memory, but I haven't verified it in any way. So if you say that `requests` rewinds the stream by its own then I guess setting `stream_file_content` should be the solution for you. So is the change fine to be merged ?

metathinker:
> 1. When setting `stream_file_content` to False SDK rollbacks to the default `requests` library behaviour. I learnt from docs that it keeps body in the memory, but I haven't verified it in any way. So if you say that `requests` rewinds the stream by its own then I guess setting `stream_file_content` to False should be the solution for you. So is the change fine to be merged ?

It's possible but I feel like in that case, calling the parameter `stream_file_content` is more confusing than helpful.

Unfortunately, I have not had time to try to plug in this change into our app to see if it works. So I would not support _or_ oppose merging this change as is.

congminh1254 (approved):
LGTM, but do we need to add docs for this?

PR #899, opened 2025-06-23
ci: Drop Python Versions 3.6 and 3.7

Hawra2020:


.github/workflows/integration-tests.yml
```diff
@@ -15,7 +15,6 @@ jobs:
       max-parallel: 1
       matrix:
         python-version:
-        - '3.6'
```
congminh1254, line 18: We should run integration test for the lowest supporting version also

setup.py
```diff
@@ -58,7 +56,7 @@ def main():
         'urllib3',
         'requests>=2.4.3,<3',
         'requests-toolbelt>=0.4.0',
-        'python-dateutil',  # To be removed after dropping Python 3.6
+        'python-dateutil',
```
congminh1254, line 59: Should we remove this dependency?

Hawra2020: Lukasz asked to keep it.

setup.py
```diff
@@ -108,6 +106,7 @@ def main():
         keywords='box oauth2 sdk',
         license='Apache Software License, Version 2.0, http://www.apache.org/licenses/LICENSE-2.0',
         package_data={'boxsdk': ['py.typed']},
+        python_requires='>=3.8',
```
mwwoda, line 109: Do we consider this a breaking change? If so, we should include exclamation point in the PR name `!` so the release notes are generated correctly - see https://github.com/box/box-node-sdk/pull/897

README.md
```diff
@@ -429,7 +429,8 @@ A current release is on the leading edge of our SDK development, and is intended
 
 | Version | Supported Environments                                  | State     | First Release | EOL/Terminated |
 |---------|---------------------------------------------------------|-----------|---------------|----------------|
-| 3       | Python 3.6+                                             | Supported | 17 Jan 2022   | TBD            |
+| 4
```
Pask423, line 432: is this change intentional, if so should we add Python 4 in integration tests ?

.github/workflows/integration-tests.yml
Pask423: +1 for @congminh1254

Pask423 (approved):

README.md
```diff
@@ -429,7 +429,7 @@ A current release is on the leading edge of our SDK development, and is intended
 
 | Version | Supported Environments                                  | State     | First Release | EOL/Terminated |
 |---------|---------------------------------------------------------|-----------|---------------|----------------|
-| 3       | Python 3.6+                                             | Supported | 17 Jan 2022   | TBD            |
+| 3       | Python 3.8+                                             | Supported | 17 Jan 2022   | TBD            |
```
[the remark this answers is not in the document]
mwwoda: @Hawra2020 Was this addressed?

congminh1254 (approved):

